PR #35, opened 2021-05-25
Feature/update and implement best practices component

Lham42:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

This change replaces the current styling of the best practices section for component documentation with a preset component. This change also adds headings to the component and reformats the list in the component.

- Every section replaced the pre-existing list with the component
- A dos and don'ts title was added for a stronger cue
- The list now has smaller bullet points whose colour matches the text
- list is aligned with the title.

**Additional context**
Add any other context here.


packages/pharos-site/src/components/statics/BestPractices.module.css
```diff
@@ -10,22 +10,38 @@
...
+
+.text_do {
+  color: var(--pharos-color-feedback-success);
+  font-weight: bold;
+}
+
+.text_dont {
+  color: var(--pharos-color-feedback-error);
+  font-weight: bold;
 }
```
daneah, line 38: There appears to be some inconsistency in the naming of these classes—some have a single underscore where others have a double underscore. I thing we've got a mix of trying to use some [BEM](http://getbem.com/introduction/) conventions but also not needing to go too wild with it since we're using CSS-in-JS and modules...either way, internal consistency might be good!

packages/pharos-site/src/components/statics/BestPractices.module.css
```diff
@@ -10,22 +10,38 @@
...
 
 .text__guideline {
   color: var(--pharos-color-marble-gray-20);
 }
+
+div div .text__guideline ul > ::marker {
+  color: var(--pharos-color-marble-gray-20);
+  font-size: 10px;
+}
```
daneah, line 47: We generally avoid using element-specific, nested, and direct descendent selectors as much as possible—I recommend adding a new class to the `ul` elements so you can target `.some-new-class ::marker` here instead.

packages/pharos-site/src/components/statics/BestPractices.module.css
```diff
@@ -10,22 +10,38 @@
...
   border: none;
   height: 2px;
   background-color: var(--pharos-color-feedback-success);
-  margin-bottom: var(--pharos-spacing-three-quarters-x);
+  margin-bottom: var(--pharos-spacing-one-half-x);
+}
+
+.text_do {
+  color: var(--pharos-color-feedback-success);
+  font-weight: bold;
```
Niznikr, line 32: We have a token for bold weight `var(--pharos-font-weight-bold)`

packages/pharos-site/src/components/statics/BestPractices.module.css
Lham42: While it still changes the sizing. The colour keeps being overwritten by the main selector and I'm not sure how to resolve that or raise the specificity of the current selector.

Niznikr (approved):

PR #50, opened 2021-05-27
fix(pharos)!: fire modal events when opened/closed via props

Niznikr:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Currently modal events are fired when they are opened/closed by a user manually. We should also fire them when the `open` prop is changed so that consumers using that method can utilize the events.

- Move `pharos-modal-opened` and `pharos-modal-closed` to updated lifecycle and remove trigger details


packages/pharos/src/components/modal/pharos-modal.ts
```diff
@@ -211,6 +211,19 @@ export class PharosModal extends LitElement {
     }
   }
 
+  private _emitEvent() {
```
daneah, line 214: Could this name be scoped a bit better? e.g. `_emitVisibilityChange` or something of the sort.

daneah (approved):

SMQuazi (approved):

michael-iden (approved):

PR #60, opened 2021-06-02
ci: add a composed Storybook build

Niznikr:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

For ease of use for consumers, we wish to include a Storybook that includes both web components and React versions of our components under a single location via [composition](https://storybook.js.org/docs/react/workflows/storybook-composition).

- Add a `main` Storybook that utilizes composition
- Add `storybook:main:start` script to run the composed Storybook locally when the `wc` and `react` ones are also running locally
- Update `build-storybook` script to build `main` Storybook and extract story info for `wc` and `react` Storybooks
- Serve `main` Storybook from https://pharos.jstor.org/storybook on the site


.storybook/main/main.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+  return {
+    webComponents: {
+      title: 'Web Components',
+      url: 'https://pharos.jstor.org/storybooks/wc',
+    },
+    react: {
+      title: 'React',
+      url: 'https://pharos.jstor.org/storybooks/react',
+    },
```
daneah, line 26: May want to end these with a slash to avoid a redirect.

daneah (approved):

.storybook/main/main.js
daneah: Er, maybe it won't even work that way given the redirect rules. Still a bit muddy for me there regarding the Netlify behavior.

PR #61, opened 2021-06-03
fix(storybook): try using more explicit patterns to fix routing

daneah:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Routing to the Storybooks still only works when a trailing slash is included. This could be because Netlify treats URLs with/without a trailing slash as equivalent to maximize cache hits, and our redirects weren't sending some requests explicitly to the backing app, instead just trying to append a slash to the incoming URL.

Create an explicit redirect to the backing app for each of:

* `/storybook`
* `/storybook/*`
* `/storybooks/react`
* `/storybooks/react/*`
* `/storybooks/wc`
* `/storybooks/wc/*`

**Additional context**
I also simplified the `from` values to be just the paths rather than the full URL, and ordered them by route from most general to most specific.

### DIFF
--- a/.changeset/proud-pumpkins-own.md
+++ b/.changeset/proud-pumpkins-own.md
@@ -1,0 +1,5 @@
+---
+'@ithaka/pharos-site': patch
+---
+
+Fix Netlify app routing involving slashes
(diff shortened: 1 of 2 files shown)


Niznikr (approved):

daneah:
Closing in favor of #62

PR #104, opened 2021-06-15
fix(pharos): update tooltip and breadcrumb overflow logic when slotted content is non-HTML

eslawski:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

When the slotted content provided to a `pharos-tooltip` is anything other than text, the desired overflow wrapping logic (30+ characters) breaks.

Example:
```html
<pharos-tooltip>
    <div>I am some text that happens to be longer than 30 characters.</div>
</pharos-tooltip>
```
 
It appears that during the [upgrade to Lit 2](https://github.com/ithaka/pharos/pull/19) we added check to only look for text nodes. This PR undoes that so we can look for nodes other than text in order to assess their lengths. Not 100% sure I understand the original context to know if this is the correct fix or not.

### DIFF
--- a/packages/pharos/src/components/tooltip/pharos-tooltip.ts
+++ b/packages/pharos/src/components/tooltip/pharos-tooltip.ts
@@ -84,9 +84,7 @@ export class PharosTooltip extends OverlayElement {
   );
 
   protected get content(): HTMLElement | Text {
-    return Array.prototype.slice
-      .call(this._contentNodes)
-      ?.find((node) => node.textContent && node.nodeName === '#text');
+    return Array.prototype.slice.call(this._contentNodes)?.find((node) => node.textContent);
   }
 
   constructor() {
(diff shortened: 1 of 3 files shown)


packages/pharos/src/components/tooltip/pharos-tooltip.ts
```diff
@@ -84,9 +84,7 @@ export class PharosTooltip extends OverlayElement {
   );
 
   protected get content(): HTMLElement | Text {
-    return Array.prototype.slice
-      .call(this._contentNodes)
-      ?.find((node) => node.textContent && node.nodeName === '#text');
+    return Array.prototype.slice.call(this._contentNodes)?.find((node) => node.textContent);
```
Niznikr, line 87:
```suggestion
    return Array.prototype.slice.call(this._contentNodes)?.find((node) => node.textContent && node.nodeName !== '#comment');
```
To ensure we don't pick up one of Lit's comment nodes

Niznikr (approved):

daneah (approved):

PR #106, opened 2021-06-15
feat(link): add min-width rule when flexed for better text wrapping support

eslawski:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

When the `pharos-link` is using the `flex` prop, the underlying link gets the `inline-flex` rule. For content that is long and containing no spaces, the flex box allows it to grow outside of its bounds. After reading [this stackoverflow post](https://stackoverflow.com/questions/36247140/why-dont-flex-items-shrink-past-content-size) I learned that applying `min-width` solves this issue and forces the content to stay within its bounds.


daneah (approved):

Lham42 (approved):

packages/pharos/src/components/link/pharos-link.scss
```diff
@@ -38,6 +38,7 @@
 
 :host([flex]) #link-element {
   display: inline-flex;
+  min-width: 0;
```
eslawski, line 41: Not sure this is having the impact I think it's having. Looking into it more...

eslawski: Alright so I think I got to the bottom of it. Check out https://github.com/ithaka/pharos/pull/106/commits/13e2cbb9ac5ad8ab58b993034c11284fa4028317

Basically needed to move the `inline-flex` rule up to the parent in order to properly apply the `min-width` like the SO post suggests. From what I can tell it appears to be behaving as we would like now. Hoping this style modification doesn't lead to any other negative consequences.

eslawski: womp chromatic seems to indicate it visually broke quite a few things

eslawski: I ended up reverting back to what I had originally. In order to make this work this work for us, as the consumer we will need to be `display: flex` on the `pharos-link` itself. This is the only way I have found to get text wrapping to work the way we want without negatively impacting links elsewhere (see previous chromatic).

Niznikr (commented):
@eslawski Thanks for the contribution! Could you include a changeset with your PR?

Niznikr (approved):

PR #123, opened 2021-06-24
documentation(site): fix react error messages for token pages

SMQuazi:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Currently, React throws messages about unique keys and reaching re-render limits. The re-render limits has to do with content being passed in as a prop. If the `content` prop is specified as a dependent, it cause an infinite loop. If it's not, animations don't get updated.

Instead of passing content as a prop, we use the built-in children prop and pass content as children elements.

**Additional context**
Unique keys were added to ColorRow to avoid duplicate keys.

### DIFF
--- a/packages/pharos-site/src/components/statics/design-token/TokenTable.tsx
+++ b/packages/pharos-site/src/components/statics/design-token/TokenTable.tsx
@@ -1,11 +1,16 @@
 import { tokenTable } from './TokenTable.module.css';
 import { useEffect, useState } from 'react';
+import type { FC } from 'react';
 
-export const TokenTable = (props: any) => {
+interface TokenTableProps {
+  title?: string;
+  subtitle?: string;
+}
+
+export const TokenTable: FC<TokenTableProps> = ({ title, subtitle, children }) => {
   const [StateTable, setStateTable] = useState(<></>);
   const Pharos =
     typeof window !== `undefined` ? require('@ithaka/pharos/lib/react-components') : null;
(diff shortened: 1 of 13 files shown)


daneah (approved):

packages/pharos-site/src/components/statics/design-token/TokenTable.tsx
```diff
@@ -1,14 +1,11 @@
 import { tokenTable } from './TokenTable.module.css';
 import { useEffect, useState } from 'react';
 
-export const TokenTable = (
-  title: string,
-  content: JSX.Element,
-  subtitle?: JSX.Element
-): JSX.Element => {
-  const [StateTable, setStateTable] = useState<JSX.Element>(<></>);
+export const TokenTable = (props: any) => {
```
Niznikr, line 4: Could you define an interface for these props and use the `FC` type for this component? That should help clear the lint warnings. You could look at some of the other site components as an example.

SMQuazi: Hmm. I think my IDE is lagging or something. It takes a while for the lint warnings to show.

Niznikr (approved):

PR #125, opened 2021-06-25
docs: add stories for sass mixins

Lham42:

- [x] Adds a new feature
- [ ] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Improves thoroughness of documentation and allows contributors to understand what mixins are used and when to apply them.

Add a storybook, with a mixin story.

**Other context**
As of now, the mixin json will have to be updated when a new mixin is added. There does not seem to be a reliable way to convert a scss into a json (particularly if we're using scss values aside from variables).


.storybook/sass.ts
```diff
@@ -0,0 +1,389 @@
+const sass_values = {
```
daneah, line 1: I worry we'll forget to update this file as we evolve our mixins. Is there a path forward that could move this documentation directly adjacent to the code of interest?

Lham42: Do you mean just moving the location or eventually having it autogenerate based on the sass file?

Niznikr: Kinda tricky as the `styles` directory contains auto generated files that we ignore. I wouldn't want to to include this file with the built package. Perhaps it can instead be put into the `docs` directory of the `pharos` package?

packages/pharos/src/styles/sass.react.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export const renderCssAttributes = (attributes) => {
+  let cssHTML = attributes.map((element, index) => {
+    if (element.indexOf(':') != -1) {
+      var property = element.substr(0, element.indexOf(':') + 1);
+      var remaining = element.substr(element.indexOf(':') + 1, element.length - 1);
+      return (
+        <div key={index} style={{ backgroundColor: '#F5F5F5' }}>
+          <code style={{ color: 'blue', backgroundColor: '#F5F5F5' }}>{property}</code>
```
Niznikr, line 30: Can we use tokens here for the colors instead?

.storybook/sass.ts
daneah: What I'm suggesting is using something like [SassDoc](http://sassdoc.com/) directly into the SASS mixin code and extracting it from there.

This can certainly be a future endeavor, but I think that'd be the ideal and would parallel the way we document the components' APIs.

Niznikr: Ah I see. Yes something like that would be ideal.

daneah: I'll file a new issue for that

daneah: That's now #127

daneah (approved):

Niznikr (approved):

packages/pharos/src/styles/sass.wc.stories.mdx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export const PharosMixins = () => html`
+  ${TokenTable(
+    'Sass Mixins',
+    html`
+      <thead>
+        <tr>
+          <th style="width:10%">Token</th>
+          <th style="width:10%">Cetegory</th>
```
Niznikr, line 61:
```suggestion
          <th style="width:10%">Category</th>
```

packages/pharos/src/styles/sass.react.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+export const PharosMixins = () => (
+  <>
+    {TokenTable(
+      'Sass Mixins',
+      <>
+        <thead>
+          <tr>
+            <th style={{ width: '10%' }}>Token</th>
+            <th style={{ width: '50%' }}>CSS attributes</th>
```
Niznikr, line 73: Missing Category here

Niznikr (commented):
The latest commit seems to introduce quite a bit of duplication. Perhaps `TokenTable` can be refactored to include the shared code?

.storybook/sass.ts
```diff
@@ -0,0 +1,426 @@
...
+        'transition-property: $properties',
+        'transition-duration: var(--pharos-transition-duration-default)',
+        'transition-timing-function: ease-in-out',
+      ],
+      usage: 'Base style for transitions',
+      category: 'Interaction',
+    },
+    {
+      name: 'content-label',
+      parameters: [':@include form-element-label($args...)'],
```
daneah, line 346: Is this one missing something? It starts with a colon.

PR #130, opened 2021-06-29
fix(modal): prevent pharos-modal-closed event from emitting upon initial render

eslawski:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

When the modal first renders the `pharos-modal-closed` event is emitted, which is incorrect. That event should only be emitted once the modal is opened THEN closed.

The `pharos-modal-closed` fires because the initial call to `updated` has all the `changedProperties` with values of `undefined`. In order to prevent this from getting into the `if` block, we now use a simple check to make sure the value is not `undefined`.


packages/pharos/src/components/modal/pharos-modal.ts
```diff
@@ -93,7 +93,7 @@ export class PharosModal extends LitElement {
   }
 
   protected updated(changedProperties: PropertyValues): void {
-    if (changedProperties.has('open')) {
+    if (changedProperties.has('open') && changedProperties.get('open') !== undefined) {
```
Niznikr, line 96: Think we only want to wrap this check around where `_emitVisibilityChange` gets called. Believe this will address the regressions from Chromatic.

eslawski: ah gotcha done!

Niznikr (approved):

Lham42 (approved):

PR #133, opened 2021-06-30
feature/image card heading level

### LINKED ISSUE #90: Remove hard-coded heading level from image cards
**The problem**
Image cards may appear at a variety of places in a document hierarchy, such that their heading level may be 2 or lower. Sometimes, it may even be desirable not to use a heading at all, instead using styled text that doesn't affect the document hierarchy.

**The solution**
Allow consumers to change the level, and potentially the tag type, for image card titles.

daneah:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Resolves #90

Add an optional `headingLevel` property that will be used for the heading level when specified. Otherwise, the previous hard-coded level will be used as a fallback.


Niznikr (approved):

jialin-he (approved):

packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -72,6 +75,13 @@ export class PharosImageCard extends LitElement {
   @property({ type: String, reflect: true, attribute: 'action-menu' })
   public actionMenu?: string;
 
+  /**
+   * The heading level to use for the card title when using the title property.
+   * @attr heading-level
```
Niznikr, line 80:
```suggestion
   * @attr heading-level
   * @type {HeadingLevel | undefined}
```
So our wrapper knows to make this optional

daneah: Looks like that's failed the tests—taking a look

Niznikr: Think you also have to export the type

`export type { HeadingLevel };`

daneah: I think it's unhappy that the component doesn't `export type HeadingLevel` (it's defined in `pharos-heading`). Can/should we `export` it from this module as well? 🤔

Niznikr: Yup!

PR #145, opened 2021-07-07
style(image-card): force 100% width for better alignment within the grid

michael-iden:

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [x] Changeset added?

Fixes issue where cards were misaligned and not taking up the full width of the grid column when their contents are narrower than the column width.

Add `width: 100%` to make cards take up full width and center their contents

**Additional context**

Before:
https://user-images.githubusercontent.com/13315416/124817457-d5514300-df37-11eb-9e40-db909dc7e22c.mov

After:
https://user-images.githubusercontent.com/13315416/124817581-fe71d380-df37-11eb-956d-5db86d8f6518.mov

### DIFF
--- a/packages/pharos/src/components/image-card/pharos-image-card.scss
+++ b/packages/pharos/src/components/image-card/pharos-image-card.scss
@@ -2,11 +2,13 @@
 
 :host {
   display: inline-flex;
+  width: 100%;
 }
 
 .card {
   display: inline-flex;
   flex-direction: column;
+  width: 100%;
 }
 
 .card__link--image {
--- a/.changeset/curvy-lies-develop.md
+++ b/.changeset/curvy-lies-develop.md
@@ -1,0 +1,5 @@
+---
+'@ithaka/pharos': patch
+---
+
+expand card to be full width to fix alignment issues for cards within a wider grid


packages/pharos/src/components/image-card/pharos-image-card.scss
```diff
@@ -2,11 +2,13 @@
 
 :host {
   display: inline-flex;
+  width: 100%;
```
michael-iden, line 5: I wasn't sure if this width should be included since we can apply it within the consuming app. But both `width: 100%` were necessary to get the right look and feel.

Niznikr (approved):

packages/pharos/src/components/image-card/pharos-image-card.scss
daneah: Adding this here in a sense makes the card a block-level element, which to me seems like the consistent thing to do based on the desired behavior.

daneah (approved):

PR #157, opened 2021-07-14
Image Card: Add `promotional` variant and `source-type` prop

### LINKED ISSUE #146: Image Card: Allow for custom aspect ratio in collection variant
**The problem**

I want to use the collection variant of the image card to get its showy heading style, but I also have images that aren't 4:3.

**The solution**

Be able to specify the aspect ratio of the image explicitly instead of [hard coding it](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/image-card/pharos-image-card.ts#L126), or better, allow the card to shrink naturally like it does for the default variant.

**Alternatives considered**

We could use the default variant to shrink to the image's natural aspect, but it doesn't have a showy heading style. We could create new images that are 4:3, but that doesn't track well.

**Additional information**

It may be that the "collection" variant is a bit of a tightly-scoped name as well—it's a bit of a "showcase" variant that may not always be showcasing a collection _per se_. That's really a semantics issue, but good to have language that can be used ubiquitously.

### LINKED ISSUE #152: Image Card: Create "promotional" variant
**The problem**
The default variant of the image card has modestly-sized title text, and the collection variant has more prominent title text. Sometimes it's useful to have even more prominent text when emphasizing a particular piece of content.

The default variant allows for images of various aspect ratios, whereas the collection variant forces images to a 4:3 aspect ratio.

**The solution**

Create a "promotional" variant (name not set in stone) that:

1. Accepts a variety of image aspect ratios
2. Uses `preset="4"` for titles
3. Requires a call to action(?)

**Alternatives considered**

This could also be handled as its own component, because it's a different content pattern from image cards at large. We don't want this to be used outside its intended context, so having a separate component might encourage a second thought about whether it's the right thing to use, compared to just being able to change the `variant` property.

**Additional information**

This may subsume #146

### LINKED ISSUE #153: Image Card: Add "source type" property
**The problem**
An image card is often used to link to content, which may itself _be_ an image, but may also be an editorial article, a journal article, a book chapter, and so on. We want a way to indicate this on the card.

**The solution**
Add a "source type" property (name not set in stone):

* Only used in the default variant
* Appears above the card title when present
* Is `pharos-color-text-40`, `pharos-font-size-small`, and `pharos-font-family-body` (default) font
* Is all caps

**Alternatives considered**

We considered the content pattern where the source type could appear below the card title, which would allow us to use the `metadata` slot to pass this in, but the information hierarchy doesn't work super well and would require more boilerplate.

daneah:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

* Resolves #152 
* Resolves #153 
* Closes #146 

* Add a `promotional` variant that behaves like the base variant but with larger font sizes for the title heading and metadata
* Add a `source-type` prop that can be used to indicate the item type of the source content


daneah:
I'm confused about the linting failure—it's not something I changed, and seems like it would've broken when the line was first introduced.

daneah:
Interesting to note: nothing yelled at me during development about the linting failure from 768a5d82593ea7e43fb06af091ff6165e083c9ac. It seems like something should've caught it earlier?

packages/pharos/src/components/image-card/pharos-image-card.scss
```diff
@@ -137,6 +149,13 @@ slot[name='metadata']::slotted(strong) {
   color: var(--pharos-color-text-20);
 }
 
+:host([variant='promotional']) .card__metadata {
+  @include mixins.font-base(
+    $font-size: var(--pharos-font-size-base),
+    $line-height: var(--pharos-line-height-medium)
+  );
```
Niznikr, line 156:
```suggestion
  @include mixins.font-base;
```
Those values are the defaults

Niznikr:
Yeah that was from `lit-analyzer` which is its own tool. I can include the ts plugin to support non VSCode IDEs and also have it scan during pre-commit.

daneah:
@Niznikr yeah I'm usually just using vim most of the time, would be appreciated 😄

packages/pharos/src/components/image-card/pharos-image-card.scss
daneah: Ah thanks—had started with different values and after changing them didn't realize.

Niznikr (approved):

jialin-he (approved):

PR #169, opened 2021-07-16
fix(select): set the selection when a value is initially passed

### LINKED ISSUE #168: Select: option should be selected if the value of select matches option value
**Expected behavior**
the value of the of the select set the selected option

**Actual behavior**
the pharos-select selected option is 1  event though the value of the select is 2

https://codesandbox.io/s/frosty-currying-rv34u?file=/src/components/HelloWorld.vue

**Pharos version**
9.x

**Your environment**
macOS
firefox

Niznikr:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Fixes #168 

- set the selection when a value is initially passed

**Additional context**
The `value` attribute is not a standard for native selects. [Frameworks such as React and Vue](https://reactjs.org/docs/forms.html#the-select-tag) use it to improve developer experience. Because our component has to render options instead of using a slot, `value` does not get reflected on initial render and so we have to manually set it.

### DIFF
--- a/packages/pharos/src/components/select/pharos-select.ts
+++ b/packages/pharos/src/components/select/pharos-select.ts
@@ -47,7 +47,7 @@ export class PharosSelect extends ObserveChildrenMixin(FormMixin(FormElement)) {
   protected firstUpdated(): void {
     if (!this.value) {
       this._setOption();
-    } else {
+    } else if (this._options.find((o) => o.value === this.value)) {
       this._select.value = this.value;
     }
     this._options.forEach((option) => (option.defaultSelected = option.hasAttribute('selected')));
--- a/packages/pharos/src/components/select/pharos-select.ts
+++ b/packages/pharos/src/components/select/pharos-select.ts
@@ -45,10 +45,10 @@ export class PharosSelect extends ObserveChildrenMixin(FormMixin(FormElement)) {
   }
 
   protected firstUpdated(): void {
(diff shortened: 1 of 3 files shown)


packages/pharos/src/components/select/pharos-select.ts
```diff
@@ -45,7 +45,11 @@ export class PharosSelect extends ObserveChildrenMixin(FormMixin(FormElement)) {
   }
 
   protected firstUpdated(): void {
-    this._setOption();
+    if (!this.value) {
+      this._setOption();
+    } else {
+      this._select.value = this.value;
+    }
```
chrisjbrown, line 52: would we want to check if `this.value` exists as a possible `option`? 
or is it kinda on the consumer at that point? 😅

Niznikr: Hmm I don't think setting `value` on the select will do anything if it doesn't exist but I can check what React does.

Niznikr: Added a check

daneah (approved):

chrisjbrown (approved):

Issue #176, opened 2021-07-22
Alert: Add ability to close


daneah:
**The problem**

Although some things are transient enough that Toast works, and some things are permanent enough that a constant alert works, there are cases where an alert should be closable.

**The solution**

Add something like a `closable` property that would render a close button that, when clicked, emits an event indicating the alert should be destroyed.

**Alternatives considered**
This content can be slotted in and done manually, but it's a usage pattern that pops up often enough that it would be best to codify it for consistency and to ensure accessibility.

**Additional context**

"Dismiss" is probably a more preferred verbiage here, but "dismissible" and "dismissable" would be hard to remember the correct spelling of, and neither is a dictionary word.

People have been doing the closing something like this:

```javascript
<pharos-alert>
    <div>
        <div>
           ...
        </div>
        <pharos-icon
            name="close"
            description="Close"
            @click="closePanel()"
        />
    </div>
</pharos-alert>
```

Where `closePanel` emits an event to the parent who reacts by hiding the alert on next render.

PR #194, opened 2021-08-06
feat: add dismissibility to alert component

Lham42:

- [x] Adds a new feature
- [] Fixes a bug
- [] Improves maintainability
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

This change allows for users to have the option to dismiss alerts.

- Add the `isDismissible` property to `PharosAlert` component.

### DIFF
--- a/packages/pharos/src/components/alert/pharos-alert.scss
+++ b/packages/pharos/src/components/alert/pharos-alert.scss
@@ -13,11 +13,12 @@
   padding: var(--pharos-spacing-1-x);
   border-radius: var(--pharos-radius-base-standard);
   display: grid;
-  grid-template-areas: 'icon body';
-  grid-template-columns: var(--pharos-spacing-one-and-a-half-x) 1fr;
+  grid-template-areas: 'icon body dismiss';
+  grid-template-columns: var(--pharos-spacing-one-and-a-half-x) 1fr 1.75rem;
   grid-gap: var(--pharos-spacing-1-x);
   background: var(--pharos-alert-color-background-base);
   color: var(--pharos-alert-color-text-base);
+  align-items: start;
 }
 
 .alert__icon {
@@ -27,6 +28,10 @@
 
 .alert__body {
   grid-area: body;
+}
+
+.alert__button {
+  grid-area: dismiss;
 }
 
 .alert--info {
--- a/packages/pharos/src/components/alert/pharos-alert.ts
+++ b/packages/pharos/src/components/alert/pharos-alert.ts
@@ -1,4 +1,4 @@
-import { LitElement, html } from 'lit';
+import { LitElement, html, nothing } from 'lit';
 import { property } from 'lit/decorators.js';
 import type { PropertyValues, TemplateResult, CSSResultArray } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
@@ -8,6 +8,7 @@ import type { PharosLink } from '../link/pharos-link';
 
 import FocusMixin from '../../utils/mixins/focus';
 import '../icon/pharos-icon';
+import '../button/pharos-button';
 
 export type AlertStatus = 'info' | 'success' | 'warning' | 'error';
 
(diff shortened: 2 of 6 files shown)


daneah (commented):
High-level feedback, based on thoughts from #176:

> Add something like a `closable` property that would render a close button that, when clicked, emits an event indicating the alert should be destroyed.
> ...
> "Dismiss" is probably a more preferred verbiage here, but "dismissible" and "dismissable" would be hard to remember the correct spelling of, and neither is a dictionary word.

I think the `is` prefix is generally not needed here either—native HTML5 boolean attributes are things like `disabled` rather than `is-disabled`.

Overall, maybe `closable` is a good alternative to `isDismissable` here?

packages/pharos/src/components/alert/pharos-alert.scss
```diff
@@ -13,11 +13,12 @@
   padding: var(--pharos-spacing-1-x);
   border-radius: var(--pharos-radius-base-standard);
   display: grid;
-  grid-template-areas: 'icon body';
-  grid-template-columns: var(--pharos-spacing-one-and-a-half-x) 1fr;
+  grid-template-areas: 'icon body dismiss';
+  grid-template-columns: var(--pharos-spacing-one-and-a-half-x) 1fr 1.75rem;
```
daneah, line 17: Should this _only_ be done when the alert is closable?

packages/pharos/src/components/alert/pharos-alert.ts
```diff
@@ -51,6 +54,13 @@ export class PharosAlert extends FocusMixin(LitElement) {
   @property({ type: String, reflect: true })
   public status!: AlertStatus;
 
+  /**
+   * Indicates if the modal is open.
```
daneah, line 58: Looks like a copy/paste error

packages/pharos/src/components/alert/pharos-alert.ts
```diff
@@ -98,6 +125,7 @@ export class PharosAlert extends FocusMixin(LitElement) {
         <div class="alert__body">
           <slot @slotchange=${this._handleSlotChange}></slot>
         </div>
+        ${this._getCloseButton()}
```
daneah, line 128: I think we've been using the `_render*` convention for functions that return HTML

packages/pharos/src/components/alert/pharos-alert.ts
```diff
@@ -27,6 +28,8 @@ const STATUSES = ['info', 'success', 'warning', 'error'];
  *
  * @slot - Contains the alert message (the default slot).
  *
+ * @fires pharos-alert-close - Fires when the toast has closed
```
Niznikr, line 31:
```suggestion
 * @fires pharos-alert-close - Fires when the alert has closed
```

Niznikr (commented):
Perhaps a test should be added to validate the button is rendered when the attribute is set?

packages/pharos/src/components/alert/pharos-alert.scss
Lham42: I guess so, or else there is the space allotted for it whether it is closable or not. Though, visually it looks the same.

daneah: I think for alerts in smaller screen real estate they now span across 1–2 additional lines of text and that text looks like it's a bit off to the left compared to before. It seems good to give them the real estate if possible, considering that most alerts at the moment are _not_ closable.

packages/pharos/src/components/alert/pharos-alert.ts
```diff
@@ -27,6 +28,8 @@ const STATUSES = ['info', 'success', 'warning', 'error'];
  *
  * @slot - Contains the alert message (the default slot).
  *
+ * @fires pharos-alert-close - Fires when the alert has closed
```
Niznikr, line 31: Looks like this needs to be fired somewhere?

daneah (commented):
A pass to replace most instances of the word "dismiss*" with "closable/close/etc." where appropriate would be welcome.

Niznikr (approved):

eslawski (approved):

PR #226, opened 2021-09-24
refactor!: require manual component registration

### LINKED ISSUE #3: Infra: Provide solution for duplicate component registration
**The problem**

Right now Pharos requires using its own `@customElement` decorator to ensure no runtime errors when a component is registered multiple times. This doesn't cover third-party components we make use of, and doesn't provide deterministic behavior if two differing versions of the component try to register (the first one wins).

**The solution**

Use the [scoped element mixin](https://github.com/open-wc/open-wc/tree/master/packages/scoped-elements) to provide an axis of freedom in registering components under another name.

**Alternatives considered**

In the long term [scoped custom element registration](https://github.com/WICG/webcomponents/issues/716) will help solve this issue natively.

Niznikr:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

Resolves #3

By default, Pharos components self-register themselves. There can be cases where multiple bundles or micro frontends utilize different versions of Pharos on the same page. Custom elements are defined on the global window namespace under the [customElements registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry). The first loaded instance gets registered. Any following registrations will throw a duplicate registration error in the browser. To mitigate this we added a decorator that simply throws a warning instead. This has made it difficult to debug and be certain the desired version of a component is actually being rendered.

To correctly solve this issue, we must remove self-registration from components and put it on the consumer to define a unique tag to use in their application. We can also use [scoped custom element registries](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md) to ensure Pharos components avoid duplicate registrations internally.

- Remove component self-registration
- Scope registries for components composed of other Pharos components using [scoped-registry-mixin](https://github.com/lit/lit/tree/main/packages/labs/scoped-registry-mixin)
- Update React wrapper to detect defined custom tag name
- Create `PharosElement` base class which sets a custom attribute `data-pharos-component` as a common identifier
- Set components to extend `PharosElement`
- Update documentation

**Additional context**
The `data-pharos-component` attribute can be used to query for a particular component regardless of their varying tag names. This can be useful for integration tests locators and unit tests.

**Migration Guidelines**

1. Remove all individual imports of Pharos web components.

2. Register them with a custom tag in your app's entrypoint:

    ```javascript
    import { PharosAlert } from '@ithaka/pharos/lib/components/alert/pharos-alert';
    customElements.define('pharos-alert-homepage', PharosAlert);
    ```

2. Add the [scoped custom element registry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry) to your app.

3. Update your bundler's production build settings to keep class names unminified:

    Webpack (Terser):

    ```javascript
    optimization: {
      minimizer: [
        new TerserPlugin({
          terserOptions: {
            keep_classnames: /^Pharos/,
            keep_fnames: /^Pharos/,
          }
        }),
      ],
    }
    ```

    Vite (ESBuild):

    ```javascript
    export default defineConfig({
      esbuild: {
        keepNames: true,
      },
    });
    ```


.changeset/quiet-suits-count.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Remove all individual imports of Pharos web components.
+
+2. Register them with a custom tag in your app's entrypoint:
+
+    ```javascript
+    import { PharosAlert } from '@ithaka/pharos/lib/components/alert/pharos-alert';
+
+    customElements.define('pharos-alert-homepage', PharosAlert);
+    ```
```
daneah, line 26: I wonder if a prefix is a better recommendation (here and in the docs), as it then acts more like a namespace. That is:

```html
<homepage-pharos-alert>
<homepage-pharos-dropdown-menu>
...
```

is easier to read/reason about than:

```html
<pharos-alert-homepage>
<pharos-dropdown-menu-homepage>
...
```

---

We could also provide a little helper (_at some point_) that could make it easier for folks who need to register a big bunch of components:

```es6
import { registerComponents } from '@ithaka/pharos/lib/utils';

import { PharosAlert } from ...
import { PharosDropdownMenu } from ...

// spits out <homepage-pharos-alert>, <homepage-pharos-dropdown-menu>, ...
registerComponents('homepage', [ PharosAlert, PharosDropdownMenu, ...])
```

The consumer could do the following, I suppose? Not sure how fun `component-name-converted-to-kebab-case` will be at first glance 🤷🏻‍♂️ 

```es6
import { PharosAlert } from ...
import { PharosDropdownMenu } from ...

[ PharosAlert, PharosDropdownMenu, ...].forEach((component) => {
    customElements.define(`homepage-${<component-name-converted-to-kebab-case>}`, component);
});
```

.changeset/quiet-suits-count.md
```diff
@@ -0,0 +1,55 @@
...
+
+2. Add the [scoped custom element registry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry) to your app.
```
daneah, line 28: Include a note to be careful to avoid double-ups (e.g. if it's already added in a more global context)?

docs/development/anatomy-of-a-component.md
```diff
@@ -153,7 +152,7 @@ import { sparklyTextStyles } from './pharos-sparkly-text.css';
 Now you've got all the ingredients needed to write the component definition. Add the following to `pharos-sparkly-text.ts`:
 
 ```typescript
-@customElement('pharos-sparkly-text')
+('pharos-sparkly-text');
 export class PharosSparklyText extends LitElement {
```
daneah, line 156: Does this need to be `PharosElement` instead?

.changeset/quiet-suits-count.md
Niznikr: No strong feeling about the convention. Curious what other folks gravitate toward.

daneah: The prefix convention further reduces the risk of collision because, when choosing a prefix, it's easier to look through existing code to understand whether that prefix is in use. As a suffix, the scannability is lost.

.changeset/quiet-suits-count.md
```diff
@@ -0,0 +1,57 @@
...
+
+    ```javascript
+    import { PharosAlert } from '@ithaka/pharos/lib/components/alert/pharos-alert';
+
+    customElements.define('homepage-pharos-alert', PharosAlert);
+    ```
+
+3. Update templates, queries, unit tests, and integration tests with the newly defined tag names.
+
+4. Add the [scoped custom element registry polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry) to a globally available location in your project.
```
NazimAli2017, line 30: Is this needed for Safari?

Niznikr: It's needed for all browsers as the proposal is not finalized yet

packages/pharos-site/gatsby-node.js
```diff
@@ -0,0 +1,26 @@
+exports.onCreateWebpackConfig = ({ getConfig, stage, actions, plugins }) => {
+  if (stage === 'build-javascript') {
+    const currentConfig = getConfig();
+
+    if (
+      currentConfig.optimization &&
+      currentConfig.optimization.minimizer &&
+      currentConfig.optimization.minimizer.length
+    ) {
```
NazimAli2017, line 9:
```suggestion
    if (currentConfig?.optimization?.minimizer.length) {
```

.changeset/quiet-suits-count.md
NazimAli2017: To prevent collisions for the same component + different versions, should we also add the version to the naming convention?

Example: globally registered `PharosAlert` on version 2.0.0, bundle registered `PharosAlert` on version 1.0.0.

NazimAli2017: 🤔  I guess it's not needed since a scoped custom element is essentially doing that with the shadow DOM.

Niznikr: Not sure I understand. Namespacing to a bundle/app will prevent any chance of collision regardless of version (even with a collision the browser will now throw an error).

NazimAli2017: Yup, I realized that after reading the specs for how the name-spacing working. Didn't realize the scoping would account for that.

NazimAli2017 (approved):
🎉 let's test this!

.changeset/quiet-suits-count.md
```diff
@@ -0,0 +1,57 @@
...
+
+1. Remove all individual imports of Pharos web components.
+
+2. Register them with a custom tag in your app's entrypoint in the form of `{app/bundle}-pharos-{component}`:
+
+    ```javascript
+    import { PharosAlert } from '@ithaka/pharos/lib/components/alert/pharos-alert';
+
+    customElements.define('homepage-pharos-alert', PharosAlert);
+    ```
```
daneah, line 26: We could also provide a little helper (_at some point_) that could make it easier for folks who need to register a big bunch of components:

```es6
import { registerComponents } from '@ithaka/pharos/lib/utils';

import { PharosAlert } from ...
import { PharosDropdownMenu } from ...

// spits out <homepage-pharos-alert>, <homepage-pharos-dropdown-menu>, ...
registerComponents('homepage', [ PharosAlert, PharosDropdownMenu, ...])
```

The consumer could do the following, I suppose? Not sure how fun `component-name-converted-to-kebab-case` will be at first glance 🤷🏻‍♂️ 

```es6
import { PharosAlert } from ...
import { PharosDropdownMenu } from ...

[ PharosAlert, PharosDropdownMenu, ...].forEach((component) => {
    customElements.define(`homepage-${<component-name-converted-to-kebab-case>}`, component);
});
```

Niznikr: The first approach does feel nice and could help minimize human error

packages/pharos/src/utils/mixins/pharos-component.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { LitElement } from 'lit';
+
+/**
+ * @param Base The base class for the mixin to modify.
+ * @returns A mixin to set a data attribute with the component class name.
+ */
+const PharosComponentImplementation = <T extends Constructor<LitElement>>(Base: T): T => {
+  /**
+   * A mixin class to handle focusing pharos components.
+   */
```
daneah, line 12: Does this need to be updated?

packages/pharos/src/utils/mixins/pharos-component.ts
```diff
@@ -0,0 +1,24 @@
+import { dedupeMixin } from '@open-wc/dedupe-mixin';
+import type { Constructor } from '@open-wc/dedupe-mixin';
+import type { LitElement } from 'lit';
+
+/**
+ * @param Base The base class for the mixin to modify.
+ * @returns A mixin to set a data attribute with the component class name.
+ */
+const PharosComponentImplementation = <T extends Constructor<LitElement>>(Base: T): T => {
```
daneah, line 9: I think this is the bit we'll want to implement ahead of the breaking change, give people a chance to upgrade to the minor version, and then all the new `[data-pharos-component]` selectors will work across the major version gap.

Niznikr: My only concern is it does require folks to update their bundler configs which may technically constitute a breaking change 🤔

daneah: Can you say more about that? Not sure I follow.

Niznikr: Step 5 of the migration guidelines is needed to prevent minification of the class names for that attribute.

daneah: I see. I think the change itself is non-breaking, but the need to update configurations is strategic and necessary before upgrading to the major version.

chrisjbrown:
I'm able to build and run with this branch
my changes are here https://github.com/ithaka/jstor/pull/7388/files

The issue i have is css... maybe
the failed "preview is not available" html/svg isn't being rendered
the date filter on the left looks off 😕 not sure why

Screen Shot 2021-10-01 at 9 44 04 AM

jialin-he:
I've got this PR with changes adopted in the workspace and it is working well!

daneah:
@chrisjbrown your issue seems to indicate that the web component is not initializing—what's showing in the screen shot seems like just the light DOM content.

PR #229, opened 2021-09-30
feat(pharos-image-card): add overlay slot

chrisjbrown:
This change adds an event emit when the user hovers over or away from the image.
This allows the consumer to react to these mouseenter/mouseleave events

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

**Additional context**
Add any other context here.


packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -161,17 +165,20 @@ export class PharosImageCard extends LitElement {
...
+        href="${this.link}"
+        label=${ifDefined(this.imageLinkLabel)}
+        subtle
+        flex
+        no-hover
+        @mouseenter=${this._handleImageHover}
+        @mouseleave=${this._handleImageHover}
+        >${this._renderLinkContent()}${this._renderHoverMetadata()}</pharos-link
+      >
+      ${this._imgHover ? html`<slot name="overlay"></slot>` : nothing}
```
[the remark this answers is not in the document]
chrisjbrown: 🤔 maybe 
one thing for my situation. 
I want to only hover when the `pharos-link` is hovered... and can't tell that without adding an eventlistener to the `pharos-link` element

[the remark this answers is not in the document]
daneah: Right—I think it makes sense to keep this model for the moment, then.

daneah (approved):

packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -127,6 +127,13 @@ export class PharosImageCard extends LitElement {
...
   private _handleImageHover(event: Event): void {
     this._title['_hover'] = event.type === 'mouseenter';
+
+    const mouseEvent = new CustomEvent(
+      event.type === 'mouseenter'
+        ? 'pharos-image-card-image-mouseenter'
+        : 'pharos-image-card-image-mouseleave'
+    );
+    this.dispatchEvent(mouseEvent);
   }
```
michael-iden, line 137: Not a blocker but I wonder if it makes sense to have 2 different handlers (one for mouse enter and other for mouse leave). Seems like that would simplify a couple conditionals without too much code duplication 🤷

michael-iden (approved):

packages/pharos/src/components/image-card/pharos-image-card.ts
chrisjbrown: yeah i could see that going both ways. 
the ` _handleImageHover` event already existed for the title hover so i just followed that path 🤔

chrisjbrown: updated. i would say it looks cleaner

Issue #253, opened 2021-11-11
Layout: gutter sizes are prohibitive and cause overflow


daneah:
**Expected behavior**

The layout grid is fluid and adjusts well to all screen sizes

**Actual behavior**

At particular screen sizes, and partially dependent on the content of a given column, the layout overflows the screen.

**Steps to reproduce the issue**

1. Visit [the layout story](https://pharos.jstor.org/storybook/?path=/story/webcomponents_components-layout--one-column)
2. Place the **Actions** panel on the right
3. Drag the panel to be wider, narrowing the area with the layout
4. Observe the layout eventually stops shrinking and overflows its portion of the screen

**Screenshots or code**

<img width="633" alt="Screen Shot 2021-11-11 at 8 52 39 AM" src="https://user-images.githubusercontent.com/1808306/141309694-2287f129-81f0-48a8-a305-d36e6cedfa68.png">

**Pharos version**

11.0.0

**Your environment**

- OS: macOS
- Browser: Chrome, Firefox, likely all
- Version: Likely all

**Additional information**

We likely want to use either smaller or fluid gutter sizes on smaller screens. Because real estate becomes precious, even a 20px gutter in a 4-column layout on a 360px screen will steal (20px * 3) / (360px) = 1/6th of the screen.

daneah:
Some [inspiration from Brad Frost](https://bradfrost.com/blog/post/layout-grid-in-design-systems/).

daneah:
We likely need to revisit this and re-develop a holistic point of view on how we handle layout systemically on the platform.

elisa-vargas:
Design (product & labs) have investigated the referenced gridless design systems recommended by Brad Frost. We agree that this approach would eliminate the current issues we are experiencing with grids. 

Design will need to create a new series of objects for the UI toolkit that will replace the current grid system. We are observing that Figma's new capabilities (e.g. setting min/max of frames; and wrapping of items within a frame) will support this work.

PR #254, opened 2021-11-11
fix(overflow): use variable-width gutters in layout

daneah:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

#253 

Use a fraction of the viewport width to determine gutter size instead of using a hard-coded value.

**Additional context**

- Add `CODEOWNERS` file for automated reviewer assignment

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ithaka/pharos-maintainers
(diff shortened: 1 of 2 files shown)


eslawski (approved):

Niznikr (approved):

daneah:
This is unlikely to be the right fix—we were able to work around this for the moment by ensuring that no content spans more columns than are available in the layout. The interaction between multiple rows of content was pretty difficult to spot and correct, so maybe some DX to consider here.

PR #267, opened 2021-12-03
feat(all): Preserve data-pharos-component attribute when Pharos class is subclassed

eslawski:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

When a Pharos component is subclassed, the subclassed name is placed in the `data-pharos-component` attribute. Instead, it would be better if the original class name was preserved.

Example:
```
class HomepagePharosButton extends PharosButton {}
```

results in `data-pharos-component=HomepagePharosButton`

This change alters the behavior so that the result is now: `data-pharos-component=PharosButton`

Adds a static `componentName` variable to each component which preserves the original pharos component name. The consumer is then free to subclass the Pharos component for scoped registry purposes.

### DIFF
--- a/packages/pharos-cli/src/templates/typescript-file-template.js
+++ b/packages/pharos-cli/src/templates/typescript-file-template.js
@@ -5,6 +5,8 @@ import type { TemplateResult, CSSResultArray } from 'lit';
 import { ${camelCaseName}Styles } from './pharos-${componentName}.css';
 
 export class Pharos${titleCaseName} extends PharosElement {
+  componentName = "Pharos${titleCaseName}";
+
   public static override get styles(): CSSResultArray {
     return [${camelCaseName}Styles];
   }
--- a/packages/pharos/README.md
+++ b/packages/pharos/README.md
@@ -29,15 +29,24 @@ $ npm install @ithaka/pharos
 
 ## Registering components
 
-1. To allow multiple versions of Pharos to exist on a page, this package only exports component classes for you to register on the [custom element registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry) in your application. To register a component, import the classes you wish to use in your application's entrypoint and define the component with a tag name in the form of `{app/bundle}-pharos-{component}`:
+1. To allow multiple versions of Pharos to exist on a page, this package only exports component classes for you to register on the [custom element registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry) in your application.
+   To register a component:
+   - Import the classes you wish to use in your application's entrypoint
+   - Declare a trivial subclass the Pharos component in the form of `{app/bundle}Pharos{component}`
+   - Define the component with a tag name in the form of `{app/bundle}-pharos-{component}`
 
 ```javascript
 import { PharosAlert } from '@ithaka/pharos/lib/components/alert/pharos-alert';
 
-customElements.define('homepage-pharos-alert', PharosAlert);
+class HomepagePharosAlert extends PharosAlert {}
+
+customElements.define('homepage-pharos-alert', HomepagePharosAlert);
 ```
 
-**Note: If you register a name that already exists the browser will throw an error about the duplicate.**
+> NOTE: Errors will occur in either of the following scenarios:
+>
+> - Define a tag name that has already been defined in the browser.
+> - Attempt to use the same `Pharos` component class to define custom elements with different names. This is especially important when the `@ithaka/pharos` dependency is shared across different apps on the page (ie. ModuleFederation).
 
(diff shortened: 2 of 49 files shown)


packages/pharos-cli/src/templates/typescript-file-template.js
```diff
@@ -5,6 +5,8 @@ import type { TemplateResult, CSSResultArray } from 'lit';
 import { ${camelCaseName}Styles } from './pharos-${componentName}.css';
 
 export class Pharos${titleCaseName} extends PharosElement {
+  componentName = "Pharos${titleCaseName}";
```
Niznikr, line 8:
```suggestion
  static componentName = "Pharos${titleCaseName}";
```
Can we not set these to `static`?

packages/pharos/README.md
```diff
@@ -29,15 +29,24 @@ $ npm install @ithaka/pharos
 
 ## Registering components
 
-1. To allow multiple versions of Pharos to exist on a page, this package only exports component classes for you to register on the [custom element registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry) in your application. To register a component, import the classes you wish to use in your application's entrypoint and define the component with a tag name in the form of `{app/bundle}-pharos-{component}`:
+1. To allow multiple versions of Pharos to exist on a page, this package only exports component classes for you to register on the [custom element registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry) in your application.
+   To register a component:
+   - Import the classes you wish to use in your application's entrypoint
+   - Declare a trivial subclass the Pharos component in the form of `{app/bundle}Pharos{component}`
+   - Define the component with a tag name in the form of `{app/bundle}-pharos-{component}`
```
daneah, line 36:
```suggestion
1. To allow multiple versions of Pharos to exist on a page, this package only exports component classes for you to register on the [custom element registry](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry) in your application. To register a component:
   - Import the classes you wish to use in your application's entrypoint
   - Declare a trivial subclass of the Pharos component in the form of `{app/bundle}Pharos{component}`
   - Define the component with a tag name in the form of `{app/bundle}-pharos-{component}`
   
As an example, the following allows an Alert to be defined for use in a home page application that will not collide with other Alerts that are defined:
```

packages/pharos/README.md
```diff
@@ -29,15 +29,24 @@ $ npm install @ithaka/pharos
...
+class HomepagePharosAlert extends PharosAlert {}
+
+customElements.define('homepage-pharos-alert', HomepagePharosAlert);
 ```
 
-**Note: If you register a name that already exists the browser will throw an error about the duplicate.**
+> NOTE: Errors will occur in either of the following scenarios:
+>
+> - Define a tag name that has already been defined in the browser.
+> - Attempt to use the same `Pharos` component class to define custom elements with different names. This is especially important when the `@ithaka/pharos` dependency is shared across different apps on the page (ie. ModuleFederation).
```
daneah, line 49:
```suggestion
> NOTE: Issues arise when you:
>
> - Define a tag name that has already been defined in the browser.
> - Attempt to use the same Pharos component class to define custom elements with different names. This is especially important when the `@ithaka/pharos` dependency is shared across different applications on the same page, for example through Webpack's module federation.
```

packages/pharos-cli/src/templates/typescript-file-template.js
eslawski: I tried, but TBH I was having some trouble getting it all to work. I could very well be missing something.

I was trying to put `static` [here in the base class](https://github.com/ithaka/pharos/blob/80bab8684d65b15dc32816fcd957e2dab9dff5dd/packages/pharos/src/utils/mixins/pharos-component.ts#L14), but it kept coming back `undefined` when I would try to do `this.dataset.pharosComponent = this.componentName;`

Then each of the components I also tried putting `static`, but it didn't seem to have any effect. It wasn't very consistent either because some components that inherit from one another like `PharosSidenavLink` required an `override` and others didn't which surprised me. I kind of figured it might have something to do with the fancy things we are doing in some components on the other side of the `extends` (ie. `ScopedRegistryMixin(...)`.

daneah (commented):
Would it be better to use `componentName = () => { this.constructor.name };` in each class? I think this would still solve the issue without allowing the actual component name and the `componentName` value to get out of sync.

daneah:
Oh, the rest just clicked. See that it might not work, unless `static` works that way—don't have enough of an environment up to test at the moment.

eslawski:
> Oh, the rest just clicked. See that it might not work, unless `static` works that way—don't have enough of an environment up to test at the moment.

@daneah,  good thought. But ya I poked around a little bit trying to get something like that to work, but haven't had any success

Niznikr:
@eslawski Think we might need to set the static field in [pharos-element.ts](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/base/pharos-element.ts) to achieve what @daneah is describing

Niznikr:
close in favor of #268

PR #281, opened 2021-12-16
ci: add size-limit action

Niznikr:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

To make package size improvements (and avoid degradations) we should have a consistent method to measure it and display this information on pull requests.

- Add [size-limit](https://github.com/ai/size-limit) package and `preset-small-lib` to measure package size via `esbuild`
- Add [GitHub action](https://github.com/andresz1/size-limit-action) that compares package size against `develop`

**Additional context**
The action correctly calculates the package size of this branch but fails because when checking out `develop` notices `size-limit` dependencies are missing. This will be fixed once merged. Can open a test PR afterwards to demonstrate how it communicates the size diff in the PR.

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -1,0 +1,38 @@
+name: Size Limit
+
+on:
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  size:
+    runs-on: ubuntu-latest
+    env:
+      CI_JOB_NUMBER: 1
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node 16.x
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - name: Get yarn cache directory path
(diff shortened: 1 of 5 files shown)


.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,38 @@
+name: Size Limit
```
daneah, line 1: Nit: My understanding is that this doesn't actually put a limit on the size right now, but rather simply reports changes in size. This might be a more apt name as a result:

```suggestion
name: Size Report
```

chrisjbrown (approved):

daneah (approved):

SMQuazi (approved):
That's pretty neat!

Issue #265, opened 2021-11-30
DropdownMenu: Disabled menu item still fires click handlers


michael-iden:
**Expected behavior**
Disabled dropdown menu item  doesn't handle any clicks. Acts like a disabled button

**Actual behavior**
Disabled button handles clicks

**Steps to reproduce the issue**

1. Add a standard dropdown to the page
2. Mark at least one menu item as disabled but add a click handler
3. Click and see it is handled

**Screenshots or code**
[CodeSandbox](https://codesandbox.io/s/unruffled-ives-h4et1?file=/public/index.html) 

**Pharos version**
11.2.1

**Your environment**

- OS: macOS
- Browser: Chrome
- Version: 96

**Additional information**
The element looks disabled when styles are included and the mouse pointer is right but functionally is still enabled.

PR #290, opened 2021-12-29
bugfix(dropdown menu item): Stop disabled dropdown menu items from triggering click events using @click attribute

satya-achanta-venkata:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

DropdownMenu: Disabled menu item still fires click handlers [#265]

Earlier, disabled dropdown menu items used to trigger click event when given @click attribute on the elements.
We found that parent element of menu element in the DOM triggering click event, so we disabled pointer for parent
element `<li>` to stop triggering click event.


packages/pharos/src/components/dropdown-menu/pharos-dropdown-menu-item.scss
```diff
@@ -68,6 +68,7 @@
   margin-right: var(--pharos-spacing-one-half-x);
 }
 
+:host([disabled]),
 :host([disabled]) .dropdown-menu-item__link,
 :host([disabled]) .dropdown-menu-item__button {
```
daneah, line 73: I believe if an element has `pointer-events: none`, its children will also have `pointer-events: none` unless otherwise specified. May be able to simplify this to simply:

```suggestion
:host([disabled]) {
```

packages/pharos/src/components/dropdown-menu/pharos-dropdown-menu-item.test.ts
```diff
@@ -110,4 +110,17 @@ describe('pharos-dropdown-menu-item', () => {
     component.dispatchEvent(event);
     expect(clickSpy.callCount).to.equal(1);
   });
+
+  it('do not trigger click event when disabled and @click handler present', async () => {
```
daneah, line 114:
```suggestion
  it('does not trigger a click event when disabled with click handler present', async () => {
```

daneah (approved):
@satya-achanta-venkata I believe the test you have here is a reasonable way to assert the behavior for the disabled state. I don't think I see another way. Adding a handler with `@click` sets up an event listener, and that means it _will_ fire, so the `preventDefault` is indeed what's needed to stop it from propagating.

packages/pharos/src/components/dropdown-menu/pharos-dropdown-menu-item.test.ts
satya-achanta-venkata: @daneah  Thank you for the clarification. Just added check for `stopPropagation` as well.

chrisjbrown (approved):

michael-iden (approved):

packages/pharos/src/components/dropdown-menu/pharos-dropdown-menu-item.scss
michael-iden, line 73: Are these two lines still needed?

daneah: I think so; the other CSS properties added underneath this rule need the specificity.

michael-iden: In my quick test using the Download MFE, it looks like we do need the `color` and `background-color` CSS applied at this level but the `pointer-events` for the `.dropdown-menu-item__link` and `.dropdown-menu-item__button` is no longer necessary if we have it at the `:host`. And the `color` and `background-color` have no effect at the `:host`. 

These rules may be the most accurate towards the effect they have. 
```
:host([disabled]) {
  pointer-events: none;
}

:host([disabled]) .dropdown-menu-item__link,
:host([disabled]) .dropdown-menu-item__button {
  color: var(--pharos-dropdown-menu-item-color-text-disabled);
  background-color: var(--pharos-dropdown-menu-item-color-background-disabled);
}
```

daneah: Agreed, these should have the equivalent effect and without mixing unlike things together.

satya-achanta-venkata: Thanks for bringing it up @michael-iden , completely Agreed. Just committed the change.

SMQuazi (approved):

PR #319, opened 2022-03-03
fix(build-styles): Enhance build-styles script to minify css

eslawski:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

The css that generated by Pharos is not minified. Specifically it's nicely formatted with newlines etc. This increases the overall size of the bundle uncessarily. If we can minify the css that's generated, we can reduce the overall size of the bundle.

The `build-styles` script is responsible for generating all the css necessary for Pharos component and putting it in the right place. It loops over each components `scss` files, runs it through a `postcss` processor, and writes it out to a file.

I believe the css was not getting minified for 2 reasons:
1. We were never using the [cssnano](https://cssnano.co/) plugin with our `postcss` processor
2. After we run it through our `postcss` processor, we format it with `prettier` before outputting it to a file, effectivley undoing any minification even if we ran it through `cssnano`. I don't know if I fully understand why we were running it through `prettier` in the first place, so interested in learning if that was there for any particular reason.


maddy531 (approved):

postcss.config.js
```diff
@@ -1,3 +0,0 @@
-module.exports = {
```
eslawski, line 1: I removed this because I don't think it was being used for `@ithaka/pharos` package since we use `postcss` directly in a script. But now I am curious if this might be used by one of the other pharos packages?

packages/pharos/src/utils/scss/_mixins.scss
```diff
@@ -340,7 +340,7 @@
 /// @access public
 /// @group links
 @mixin link-base {
-  @include font-base($font-size: inherit, $line-height: inherit);
+  @include font-base($line-height: inherit);
```
eslawski, line 343: This fixes an issue that blows up when you try to run it through `cssnano`. I think it actually generates invalid css. We were passing `$font-size: inherit` to `font-base`. This resulted in a css rule like:

`letter-spacing: calc(inherit * -.02)`

What's more interesting is that we have a `letter-spacing` rule a few lines down in this file anyways. So it get's overriden either way.

daneah: I'm wary of this change. I think it might be what's impacting the letter spacing we see in the diff of the Chromatic build.

eslawski: Ya that is peculiar. Will dig in more

daneah: I'd think a conditional around the `letter-spacing` rule if `$font-size` is `inherit` might do the trick

postcss.config.js
Niznikr: Believe it was used for the Storybook build.

eslawski: ah gotcha @Niznikr . Added it back.

daneah (approved):

satya-achanta-venkata (approved):

PR #330, opened 2022-03-15
feature: add optional horizontal line to tabs

NazimHAli:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Makes it easier for users to evaluate tabs visually

Adds a boolean property to the tabs component.

### DIFF
--- a/packages/pharos/src/components/tabs/pharos-tabs.test.ts
+++ b/packages/pharos/src/components/tabs/pharos-tabs.test.ts
@@ -244,4 +244,27 @@ describe('pharos-tabs', () => {
     await component.updateComplete;
     expect(count).to.equal(0);
   });
+
+  it('does not render horizontal line', async () => {
+    const spanElement = component.renderRoot.querySelector('span') as HTMLDivElement;
+
+    expect(spanElement).to.be.null;
+  });
+
+  it('renders horizontal line', async () => {
+    component = await fixture(html`
+      <pharos-tabs horizontalline="true">
+        <pharos-tab id="tab-1" data-panel-id="panel-1">Tab 1</pharos-tab>
+        <pharos-tab id="tab-2" data-panel-id="panel-2">Tab 2</pharos-tab>
+        <pharos-tab id="tab-3" data-panel-id="panel-3">Tab 3</pharos-tab>
+        <pharos-tab-panel id="panel-1" slot="panel">Panel 1</pharos-tab-panel>
+        <pharos-tab-panel id="panel-2" slot="panel">Panel 2</pharos-tab-panel>
+        <pharos-tab-panel id="panel-3" slot="panel">Panel 3</pharos-tab-panel>
+      </pharos-tabs>
+    `);
+
+    const spanElement = component.renderRoot.querySelector('span') as HTMLDivElement;
+
+    expect(spanElement).not.to.be.null;
+  });
 });
--- a/packages/pharos/src/components/tabs/pharos-tabs.test.ts
+++ b/packages/pharos/src/components/tabs/pharos-tabs.test.ts
@@ -253,7 +253,7 @@ describe('pharos-tabs', () => {
 
   it('renders panel separator', async () => {
     component = await fixture(html`
-      <pharos-tabs panelseparator="true">
+      <pharos-tabs panel-separator>
         <pharos-tab id="tab-1" data-panel-id="panel-1">Tab 1</pharos-tab>
         <pharos-tab id="tab-2" data-panel-id="panel-2">Tab 2</pharos-tab>
         <pharos-tab id="tab-3" data-panel-id="panel-3">Tab 3</pharos-tab>
--- a/packages/pharos/src/components/tabs/pharos-tabs.test.ts
+++ b/packages/pharos/src/components/tabs/pharos-tabs.test.ts
@@ -245,15 +245,15 @@ describe('pharos-tabs', () => {
     expect(count).to.equal(0);
   });
 
-  it('does not render horizontal line', async () => {
-    const spanElement = component.renderRoot.querySelector('span') as HTMLDivElement;
+  it('does not render panel separator', async () => {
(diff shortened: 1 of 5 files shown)


chrisjbrown (approved):

packages/pharos/src/components/tabs/pharos-tabs.test.ts
```diff
@@ -244,4 +244,27 @@ describe('pharos-tabs', () => {
...
+
+  it('does not render horizontal line', async () => {
+    const spanElement = component.renderRoot.querySelector('span') as HTMLDivElement;
+
+    expect(spanElement).to.be.null;
+  });
+
+  it('renders horizontal line', async () => {
+    component = await fixture(html`
+      <pharos-tabs horizontalline="true">
```
daneah, line 256: It feels like e.g. `panel-separator` might provide more clarity here. "Horizontal line" leaves a fair amount to the imagination!

packages/pharos/src/components/tabs/pharos-tabs.test.ts
```diff
@@ -244,4 +244,27 @@ describe('pharos-tabs', () => {
...
+        <pharos-tab id="tab-1" data-panel-id="panel-1">Tab 1</pharos-tab>
+        <pharos-tab id="tab-2" data-panel-id="panel-2">Tab 2</pharos-tab>
+        <pharos-tab id="tab-3" data-panel-id="panel-3">Tab 3</pharos-tab>
+        <pharos-tab-panel id="panel-1" slot="panel">Panel 1</pharos-tab-panel>
+        <pharos-tab-panel id="panel-2" slot="panel">Panel 2</pharos-tab-panel>
+        <pharos-tab-panel id="panel-3" slot="panel">Panel 3</pharos-tab-panel>
+      </pharos-tabs>
+    `);
+
+    const spanElement = component.renderRoot.querySelector('span') as HTMLDivElement;
```
daneah, line 266: It might be best to be specific about the element here, using the class (whatever it ends up being per my other naming suggestion)

```suggestion
    const spanElement = component.renderRoot.querySelector('.horizontal-line') as HTMLDivElement;
```

packages/pharos/src/components/tabs/pharos-tabs.test.ts
```diff
@@ -244,4 +244,27 @@ describe('pharos-tabs', () => {
...
+
+  it('does not render panel separator', async () => {
+    const spanElement = component.renderRoot.querySelector('.panel-separator') as HTMLDivElement;
+
+    expect(spanElement).to.be.null;
+  });
+
+  it('renders panel separator', async () => {
+    component = await fixture(html`
+      <pharos-tabs panelseparator="true">
```
daneah, line 256: Does this also work with `panel-separator` as two words? That would be my expectation so want to make sure I'm not missing anything.

packages/pharos/src/components/tabs/pharos-tabs.ts
```diff
@@ -16,6 +17,13 @@ import type { PharosTabPanel } from './pharos-tab-panel';
  *
  */
 export class PharosTabs extends PharosElement {
+  /**
+   * If should display a panel separator.
+   * @attr panelSeparator
+   */
+  @property({ type: Boolean, reflect: true })
+  public panelSeparator = false;
+
```
Niznikr, line 26:
```suggestion
  /**
   * If should display a panel separator.
   * @attr panel-separator
   */
  @property({ type: Boolean, reflect: true, attribute: 'panel-separator' })
  public panelSeparator = false;

```

packages/pharos/src/components/tabs/pharos-tabs.test.ts
Niznikr, line 256:
```suggestion
      <pharos-tabs panel-separator>
```
For boolean props, the presence of its attribute is truthy

packages/pharos/src/components/tabs/pharos-tabs.wc.stories.mdx
```diff
@@ -87,3 +87,26 @@ import { GuidelineLink } from '@config/GuidelineLink';
     `}
   </Story>
 </Canvas>
+
+## Panel Separator
+
+<Canvas>
+  <Story name="Panel Separator">
+    {html`
+      <pharos-tabs panelseparator="true" style="width: 100%">
```
Niznikr, line 96:
```suggestion
      <pharos-tabs panel-separator style="width: 100%">
```

packages/pharos/src/components/tabs/pharos-tabs.test.ts
daneah: Ah, I think @Niznikr's suggestion below does the trick.

NazimHAli: Is this preferred? I added `true` so that it was verbose.

daneah: Strictly speaking, `true` is not an allowed value—the value must be one of:

* unassigned
* empty string
* a case-sensitive match for the attribute name itself

See more on the MDN docs [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes).

Niznikr: Yes since otherwise it may be misleading for folks and they might think they could do `panel-separator="false"` which is not the case.

NazimHAli: Wow, boolean attributes have different rules than I assumed. Thanks for that info!

daneah (approved):

Niznikr (approved):

PR #366, opened 2022-07-07
ImageCard: Selectable and disabled variants, new styling for error variant

ymouzakis:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

This PR adds three new features to the image card:

1. New `selectable` and `collection-selectable` image card variants. There is also a `subtle-select` prop that more discretely shows the card's selectability.
2. New image card 'disables' states
3. New styling for the error state

All three of these (and some of their combinations) are visible here:

Screen Shot 2022-07-07 at 4 52 51 PM

- Added new selectable variants to image card
- Added props for disabled and sublte-select
- Used variants and props to render checkbox
- Added onMouseEnter/Exit listeners to handle those events
- Use events, props, and classMap to use relevant classes for each variant's events

**Additional context**
Add any other context here.

### DIFF
--- a/packages/pharos/src/components/image-card/pharos-image-card.scss
+++ b/packages/pharos/src/components/image-card/pharos-image-card.scss
@@ -21,7 +21,7 @@
   background-color: var(--pharos-color-marble-gray-94);
 }
 
-.card__strong-selectable-hover {
+.card__selectable-card-hover {
   background-color: var(--pharos-color-marble-gray-base);
 }
 
--- a/packages/pharos/src/components/image-card/pharos-image-card.scss
+++ b/packages/pharos/src/components/image-card/pharos-image-card.scss
@@ -21,6 +21,10 @@
   background-color: var(--pharos-color-marble-gray-94);
 }
 
+.card__strong-selectable-hover {
+  background-color: var(--pharos-color-marble-gray-base);
+}
+
 slot[name='image']::slotted(img) {
   height: 100%;
   width: 100%;
@@ -31,6 +35,7 @@ slot[name='image']::slotted(img) {
 .card__selected {
   background-color: var(--pharos-color-white);
   border: solid 2px var(--pharos-color-interactive-secondary);
+  box-sizing: border-box;
 
   slot[name='image']::slotted(img) {
     height: calc(100% - 2 * var(--pharos-spacing-three-quarters-x));
--- a/packages/pharos/src/components/image-card/pharos-image-card.scss
+++ b/packages/pharos/src/components/image-card/pharos-image-card.scss
@@ -14,7 +14,11 @@
 .card__link--image {
   position: relative;
   height: 14rem;
-  margin: 0 auto var(--pharos-spacing-1-x);
+  justify-content: center;
+}
+
+.card__selected {
+  background-color: var(--pharos-color-marble-gray-94);
 }
 
 .card__link--image,
@@ -109,6 +113,9 @@
 
 .card__selector {
   position: absolute;
+  z-index: 1;
+  margin-top: var(--pharos-spacing-one-quarter-x);
+  margin-left: var(--pharos-spacing-one-quarter-x);
 }
 
 :host(:not([variant='promotional'])) {
--- a/packages/pharos/src/components/image-card/pharos-image-card.scss
+++ b/packages/pharos/src/components/image-card/pharos-image-card.scss
@@ -27,7 +27,7 @@
 
 .card__selected {
   background-color: var(--pharos-color-white);
-  border: solid 2px var(--pharos-color-interactive-secondary);
+  outline: solid 2px var(--pharos-color-interactive-secondary);
   box-sizing: border-box;
   transition: background-color var(--pharos-transition-duration-short) ease-in-out;
 
@@ -39,7 +39,7 @@
 }
 
 :host([disabled]) .card__selected {
-  border: solid 2px var(--pharos-color-marble-gray-50);
+  outline: solid 2px var(--pharos-color-marble-gray-50);
 }
 
 :host([disabled]) .card__link--image {
@@ -143,7 +143,7 @@ slot[name='image']::slotted(img) {
 .card__link--collection {
   display: grid;
   position: relative;
-  height: 9rem;
+  height: auto;
   margin-bottom: var(--pharos-spacing-1-x);
 }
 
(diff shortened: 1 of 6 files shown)


packages/pharos/src/components/image-card/pharos-image-card.wc.stories.mdx
```diff
@@ -123,6 +123,35 @@ export const PromotionalTemplate = () => html`<pharos-layout style="margin: 1rem
   </div>
 </pharos-layout>`;
 
+export const Selectable = () => html`<pharos-layout tag="ol" style="margin: 1rem 0">
```
ymouzakis, line 126: I will trim this file down before merge. I wanted to temporarily create all combinations of props/variants to prove it was working.

packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -245,9 +324,63 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
...
+        ${this.renderTitle} ${this._renderActionButton()}
+      </div>
       ${this._renderMetadata()}
     </div>`;
   }
+
+  private _cardToggleSelect(event: Event): void {
+    if (this._isSelectableViaCard() || (event.target as Element)?.nodeName == 'PHAROS-CHECKBOX') {
+      // this is required to prevent navigation on the link click
+      event.preventDefault();
```
ymouzakis, line 342: I think would be better to remove the links from the card title and thumbnail when cards can be selected instead of overriding the link....

ymouzakis: I'm building confidence that leaving the links as is and using `event.preventDefault();` to avoid the link navigation is a good way to go. I feel that way because we want the styling and red on hover styling a link provides. Still not 100% convinced though. Very open to suggestions here!

packages/pharos/src/components/image-card/pharos-image-card.scss
```diff
@@ -133,13 +210,6 @@
   }
 }
 
-slot[name='image']::slotted(img) {
```
ymouzakis, line 136: The linter was complaining that this needed to move up in the file..

ymouzakis:
I still need to write tests and make a change set.

packages/pharos/src/components/image-card/pharos-image-card.wc.stories.mdx
```diff
@@ -280,519 +305,83 @@ export const Selectable = () => html`<pharos-layout tag="ol" style="margin: 1rem
...
-        <strong id="items-${index}" slot="metadata">${collection.items} items</strong>
-        <div id="description-${index}" slot="metadata">
-          Selections from the global permanent collection.
-        </div>
-      </pharos-image-card>
-    </li>`;
-  })}</pharos-layout
->`;
+  <li class="image-card-example__card--collection">
+    <pharos-image-card title="Selectalbe" link="#" variant="selectable-collection">
```
daneah, line 319:
```suggestion
    <pharos-image-card title="Selectable" link="#" variant="selectable-collection">
```

packages/pharos/src/components/image-card/pharos-image-card.scss
```diff
@@ -120,6 +178,32 @@
...
+
+.card__container-selected--error {
+  background-color: var(--pharos-color-white);
+  transition: background-color var(--pharos-transition-duration-short) ease-in-out;
+}
+
+.card__container-selectable-hover--error {
+  background-color: inherit;
+  transition: background-color var(--pharos-transition-duration-short) ease-in-out;
+}
```
daneah, line 205: This SCSS isn't quite following the existing BEM approach the project uses. Based on the way this is all being used I think something like the following is more in line, requiring similar changes in the DOM:

```scss
.card__link--image.card__link--selectable {
  background-color: var(--pharos-color-marble-gray-94);
}

.card--selectable.card--hover {
  background-color: var(--pharos-color-marble-gray-base);
  transition: background-color var(--pharos-transition-duration-short) ease-in-out;
}

.card__container--error.card__container--selected {
  background-color: var(--pharos-color-white);
  transition: background-color var(--pharos-transition-duration-short) ease-in-out;
}

.card__container--error.card__container--selectable.card__container--hover {
  background-color: inherit;
  transition: background-color var(--pharos-transition-duration-short) ease-in-out;
}
```

packages/pharos/src/components/image-card/pharos-image-card.scss
```diff
@@ -24,6 +25,52 @@
...
+  outline: solid 2px var(--pharos-color-interactive-secondary);
+  box-sizing: border-box;
+  transition: background-color var(--pharos-transition-duration-short) ease-in-out;
+
+  slot[name='image']::slotted(*) {
+    height: calc(100% - 2 * var(--pharos-spacing-three-quarters-x));
+    width: calc(100% - 2 * var(--pharos-spacing-three-quarters-x));
+    padding: var(--pharos-spacing-three-quarters-x);
+  }
+}
```
daneah, line 39: This is currently being used as a bit of a utility class, which doesn't follow BEM very well; consider creating e.g. a `.card__link--collection.card__link--selected` and `.card__link--image.card__link--selected` split of this.

daneah (commented):
Overall this is looking really good; I have a couple of suggestions on improving the CSS with an eye toward BEM.

packages/pharos/src/components/image-card/pharos-image-card.scss
daneah: The BEM philosophy takes a different approach: classes should be used to describe component structure and variations. Some of these variations may have the same _outcome_, but are achieved through applying styles in the context of the specific component. One way to achieve both what I'm saying and what you're saying is to use a [mixin](https://sass-lang.com/documentation/at-rules/mixin) or a [placeholder/extension](https://thesassway.com/understanding-placeholder-selectors/) to reduce the repeated CSS.

ymouzakis: Thanks for the explanation. The mixin is working nice.

daneah (approved):
Thanks for all the work on this; it's looking pretty good to me! I'd like to get one more review from the team.

daneah (changes requested):
@ymouzakis Please add a changeset—then this will be good to go.

daneah (approved):

Issue #377, opened 2022-07-19
ImageCard: Subtle select is not selectable when using Safari


michael-iden:
**Expected behavior**
Subtle select cards are selectable when using Safari

**Actual behavior**
Cannot select the subtle cards in Safari

**Steps to reproduce the issue**

1. Go to https://pharos.jstor.org/storybook/?path=/story/webcomponents_components-image-card--selectable in Safari
2. Hover over the second card to make it selectable
3. Click checkbox

**Screenshots or code**
https://pharos.jstor.org/storybook/?path=/story/webcomponents_components-image-card--selectable 

**Pharos version**
12.5.1

**Your environment**

- OS: macOS
- Browser: Safari
- Version: 15.5

**Additional information**
Works in other browsers

PR #384, opened 2022-07-29
fix(card): restructure cards to avoid extra mouse events preventing selection in Safari

michael-iden:

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

#377 

Moves the checkbox within the rest of the card element where it is visually contained
Remove mouse listener events on checkbox

**Additional context**
This also looks to remove re-render on mouse move when hovering the checkbox. If mouse is moving a lot that could add up to hundreds to thousands of re-renders in a short period of time. 

https://user-images.githubusercontent.com/13315416/181800423-a9c49d9b-685b-483a-94ab-358c96391e37.mov

### DIFF
--- a/packages/pharos/src/components/image-card/pharos-image-card.ts
+++ b/packages/pharos/src/components/image-card/pharos-image-card.ts
@@ -183,6 +183,7 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
     if (!this.disabled) {
       this._title['_hover'] = true;
       this._handleMouseEnterSelectable();
+
       const mouseEvent = new CustomEvent('pharos-image-card-image-mouseenter');
       this.dispatchEvent(mouseEvent);
     }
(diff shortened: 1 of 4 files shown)


packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -404,8 +418,6 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
           ?checked=${this._isSelected}
           ?disabled=${this.disabled}
           name="Select ${this.title}"
-          @mouseenter=${this._handleMouseEnterSelectable}
-          @mouseleave=${this._handleMouseLeaveSelectable}
```
michael-iden, line 408: These mouseevents and the moving of the `${this._renderCheckbox()}` in the templates are the focus of the fix.

ymouzakis: I think these changes look really good. Did you have a chance to yalc the changes into Workspace? I think it's fine, but we may want to confirm that the folder cards still look the same.

daneah (approved):

packages/pharos/src/components/image-card/pharos-image-card.ts
michael-iden: I checked it out in Chrome and looks good. Safari can be a little tricky/tedious with the cookie mapping to test out yalc'd over changes locally.

ymouzakis (approved):

PR #390, opened 2022-08-16
feat(site): add last updated column to status page

daneah:

- [x] Adds a new feature
- [ ] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

We would like a quick reference to understand when a component was last changed.

- Remove obviated Callout row
- Add last updated versions to status.json
- Link version text to relevant changelog heading (this may not work for
  all versions as some changelog entries are in a legacy file)

**Additional context**

This will not be ideal to update, at the moment, but gives us a starting place. The main challenge being that, at the time that a change is merged, the release version number is not necessarily known. Multiple changes may be included in a release, and even though one change may be a patch another change may be a major bump, meaning the release version will be a major bump. One mitigation strategy is to release every change no matter how small; this has the added advantage of making all patch and minor changes available to consumers without having to deal with a major version upgrade, but also increases release management burden. Given the small volume of releases we have currently, this may be fine.

Screen Shot 2022-08-16 at 14 54 05

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,6 +17,7 @@
 - [ ] Test suite(s) passing?
 - [ ] Code coverage maximal?
 - [ ] Changeset added?
+- [ ] Component status page up to date?
 
 **What does this change address?**
 A clear and concise description or a direct link to an issue [e.g. #15]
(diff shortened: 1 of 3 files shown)


fswlee:
this makes sense to me @daneah !

satya-achanta-venkata (approved):

jialin-he (approved):

PR #393, opened 2022-08-17
ImageCard: remove event.stopPropagation from _cardToggleSelect

acjreno:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Allows the click event on image cards to propagate in order to close modals on the same page.

Removed event.stopPropagation from _cardToggleSelect, and added a check in order to prevent the main body of the function from being executed twice.

### DIFF
--- a/packages/pharos/src/components/image-card/pharos-image-card.ts
+++ b/packages/pharos/src/components/image-card/pharos-image-card.ts
@@ -364,12 +364,13 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
   private _cardToggleSelect(event: Event): void {
     if (
       !this.disabled &&
-      (this._isSelectableViaCard() || (event.target as Element)?.nodeName == 'PHAROS-CHECKBOX')
+      ((this._isSelectableViaCard() && (event.target as Element)?.nodeName != 'PHAROS-CHECKBOX') ||
+        (event.currentTarget as Element)?.nodeName == 'PHAROS-CHECKBOX')
     ) {
       // this is required to prevent navigation on the link click
       event.preventDefault();
-      event.stopPropagation();
       this._isSelected = !this._isSelected;
+
       this.dispatchEvent(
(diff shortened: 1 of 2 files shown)


michael-iden (approved):
Changes LGTM. Not positive what the test failure is about, especially since it is firefox only

packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -364,12 +364,13 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
   private _cardToggleSelect(event: Event): void {
     if (
       !this.disabled &&
-      (this._isSelectableViaCard() || (event.target as Element)?.nodeName == 'PHAROS-CHECKBOX')
+      ((this._isSelectableViaCard() && (event.target as Element)?.nodeName != 'PHAROS-CHECKBOX') ||
+        (event.currentTarget as Element)?.nodeName == 'PHAROS-CHECKBOX')
```
daneah, line 368: This is of the form `(A ^ !B) v B)` which is equivalent to `A v B`:

```suggestion
      (this._isSelectableViaCard() ||
        (event.currentTarget as Element)?.nodeName == 'PHAROS-CHECKBOX')
```

michael-iden: I feel like there might be something a little extra going on because of the event bubbling up. When the checkbox is clicked and handles the event here, which enables the card selection and `_isSelectableViaCard()` becomes `true`. 

Then as the event bubbles, the card handles the event and since `_isSelectableViaCard` is `true` it gets handled again and we deselect the card.

daneah (approved):

PR #408, opened 2022-09-19
ImageCard: add selectable treatment to subtle cards

michael-iden:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

If the card has `subtle` set to `true`, the metadata shows on hover but the checkbox would be covered up and overlapping. This moves the checkbox to be within the metadata hover overlay for subtle cards, above the title.

- Changes when we put the checkbox within the DOM for subtle cards
- Add classes and update when they are applied for proper positioning

**Additional context**
The card is getting somewhat difficult to maintain with the list of variants and also additional boolean attributes that can also be treated like a variant. `subtle` also used in two different contexts within in the component. I'd be open to a discussion around to both clean up the interface/implementation and work with design to limit the number of card experiences we support

image


daneah (approved):
This looks great. I see your concern about the complexity of the component; I can raise this in the next Pharos operations discussion.

packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -254,7 +254,9 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
   private _renderHoverMetadata(): TemplateResult | typeof nothing {
     return this.subtle
       ? html`<div class="card__metadata--hover">
-          <strong class="card__title--hover">${this.title}</strong><slot name="metadata"></slot>
+          ${this._renderCheckbox()}
+          <strong class="card__title--hover">${this.title}</strong>
+          <slot name="metadata"></slot>
```
jialin-he, line 259: Do we need to hide the metadata when the card is selected?

michael-iden: AFAIK yes, although I haven't had a chance to run this by Lori yet to ensure its looking as she expected

daneah:
@michael-iden you might already be on it, but it looks like the latest commits broke some image displays and heading styles.

packages/pharos/src/components/image-card/pharos-image-card.wc.stories.mdx
```diff
@@ -177,6 +177,70 @@ export const Selectable = () => html`<pharos-layout tag="ol" style="margin: 1rem
   </li>
 </pharos-layout>`;
 
+export const SubtleSelectable = () => html`<pharos-layout tag="ol" style="margin: 1rem 0">
```
daneah, line 180: This appears to be an exact copy of the existing *Selectable* story which already includes the subtle selectable card—might we not need this?

packages/pharos/src/components/image-card/pharos-image-card.ts
michael-iden: Did give Lori a moment to look a this and you are right. We do now hide the metadata on selection

https://user-images.githubusercontent.com/13315416/191797133-d84c0f24-8ac7-49ee-9918-c0669d15dbff.mov

packages/pharos/src/components/image-card/pharos-image-card.wc.stories.mdx
michael-iden: It's a little different in that this block has the `subtle="true"` on all the card elements, whereas the `Selectable` story does not. But I think that reinforces the confusion over the # of variations we have here and the difficulty supporting them.

image

jialin-he (approved):

PR #409, opened 2022-09-20
Link: Allow for visited links styling

rehanabbasi:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

We want the visited links styling to show up for Pharos links, when the user has visited the link. The consumer of the component should be able to either apply the visited links styling or not apply by specifying the `visitable` property on the Pharos Link component, which is defaulted to `false`

 We add css rule to apply the required styling for the pharos link instance that has `visitable` property specified as `true`,

### DIFF
--- a/packages/pharos/src/components/link/pharos-link.scss
+++ b/packages/pharos/src/components/link/pharos-link.scss
@@ -95,6 +95,11 @@
   color: var(--pharos-color-interactive-secondary);
 }
 
+:host([visitable]) #link-element:visited {
+  color: var(--pharos-color-night-blue-base);
+  text-decoration-color: var(--pharos-color-night-blue-base);
+}
+
 :host([no-hover][subtle]) #link-element:hover {
   @include mixins.no-underline;
 }
--- a/packages/pharos/src/components/link/pharos-link.scss
+++ b/packages/pharos/src/components/link/pharos-link.scss
@@ -9,8 +9,16 @@
   @include mixins.link-hover;
   @include mixins.interactive-focus;
 
-  &:active {
+  @at-root &:active {
     @include mixins.underline($text-decoration-color: var(--pharos-color-hover-primary));
+  }
+
+  @at-root &.link--alert:hover {
+    color: var(--pharos-color-interactive-secondary);
+  }
+
+  @at-root &.link--alert:active {
+    @include mixins.underline;
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/pharos/src/components/link/pharos-link.scss
```diff
@@ -95,6 +95,11 @@
   color: var(--pharos-color-interactive-secondary);
 }
 
+:host([visitable]) #link-element:visited {
```
daneah, line 98: The name visitable implies that other links without this attribute are **not** visitable, which is probably a source of confusion. I discussed this previously with @chrisjbrown but it looks like we didn't reach a conclusion on another name; might we call this something like `indicate-visited`?

packages/pharos/src/components/link/pharos-link.ts
```diff
@@ -77,6 +77,13 @@ export class PharosLink extends FocusMixin(AnchorElement) {
   @property({ type: Boolean, reflect: true, attribute: 'no-hover' })
   public noHover = false;
 
+  /**
+   * Indicates if the link should render the visited link styling.
+   * @attr bold
```
daneah, line 82: Be sure to update this.

daneah (commented):
@rehanabbasi Please also add a story to the Storybook that contains a link to some external site so people can see how visited links will display. I guess we can't force it to display as visited in Storybook (I think) for automated testing, but people can at least confirm it manually.

packages/pharos/src/components/link/pharos-link.scss
chrisjbrown: yeah we didn't land on a different name
`indicate-visited` sounds good to me

packages/pharos/src/components/link/pharos-link.ts
rehanabbasi: Updated in [ab9058a](https://github.com/ithaka/pharos/pull/409/commits/ab9058a239f6a6b1f2ce37a40e461ca48679f1a9)

packages/pharos/src/components/link/pharos-link.scss
rehanabbasi: Updated in [ab9058a](https://github.com/ithaka/pharos/pull/409/commits/ab9058a239f6a6b1f2ce37a40e461ca48679f1a9)

daneah:
@rehanabbasi I'm looking at [the story](https://60919c26122bd50039b34644-owlrlfnkhs.chromatic.com/?path=/story/components-link--visited-link) and it seems like the link goes through an awkward transition on hover when it has visited styling. It starts out night blue but then very briefly transitions to black before transitioning to red, and then does the reverse when you stop hovering. This indicates some disagreeing transition properties and should be fixed IMO. @fswlee I'd appreciate your input to see if you concur.

fswlee:
@daneah I don't see the transition you are describing. This is what I see! 

[Components-Link---Visited-Link-⋅-Storybook.webm](https://user-images.githubusercontent.com/5896534/191539101-426ce5c5-616d-48ab-b186-1d9749c2d1d4.webm)

fswlee:
@rehanabbasi I see what Dane is referring to now. Each browser seems to be doing something different and unexpected.

On hover, we want to transition to the red with no underline.

rehanabbasi:
@daneah @fswlee the fix for that color transition issue is made in [f25857f](https://github.com/ithaka/pharos/pull/409/commits/f25857fc990d36aed333aa401a8ad88be8e063e0) commit. I was able to see it happening on firefox, but should now be fixed.

daneah:
@rehanabbasi I don't think I agree; I still see a momentary flash of black before it transitions to red.

daneah:
@rehanabbasi it's possible you're running into some issues with the order of definition for the rules. See [this Twitter thread](https://twitter.com/meyerweb/status/1279053893440282625?s=20&t=h50JxOFLAU-mMN9DizegNw) about the ordering.

daneah:
@rehanabbasi it appears what I'm seeing is a manifestation of [this Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=868975) that is still open after 9 years. Because this has low impact I'm inclined for us to move forward. I guess that I'm thankful it uncovered some Safari display issues in the meantime and that we were able to fix those up 😅

daneah (approved):

PR #426, opened 2022-10-14
chore: pin Yarn version using Yarn's built-in version managment

drewgingerich:
This PR updates Yarn configuration to use the current standard Yarn configuration file format. It also embeds and pins the Yarn version to ensure all developers use the same Yarn version by default. This should improve the reproducibility of the dev environment for developers that are new to the project.


drewgingerich:
Closing this in favor of a future PR with a few more DX changes

PR #431, opened 2022-10-24
Checkbox: only update state if change was not cancelled

chrisjbrown:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

**Additional context**
Add any other context here.

### DIFF
--- a/packages/pharos/src/components/checkbox/pharos-checkbox.ts
+++ b/packages/pharos/src/components/checkbox/pharos-checkbox.ts
@@ -64,15 +64,17 @@ export class PharosCheckbox extends FormMixin(FormElement) {
   }
 
   public onChange(): void {
-    this.indeterminate = false;
-    this.checked = this._checkbox.checked;
-
-    this.dispatchEvent(
+    const notCancelled = this.dispatchEvent(
       new Event('change', {
         bubbles: true,
         composed: true,
       })
     );
+
+    if (notCancelled) {
+      this.indeterminate = false;
+      this.checked = this._checkbox.checked;
+    }
   }
 
   _handleFormdata(event: CustomEvent): void {


packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -64,15 +64,17 @@ export class PharosCheckbox extends FormMixin(FormElement) {
...
   public onChange(): void {
-    this.indeterminate = false;
-    this.checked = this._checkbox.checked;
-
-    this.dispatchEvent(
+    const notCancelled = this.dispatchEvent(
       new Event('change', {
         bubbles: true,
         composed: true,
       })
```
daneah, line 71: Does this need `cancelable: true`?

chrisjbrown: oh... yes it does

daneah:
@chrisjbrown that's a good point about the order of events leading to a change in behavior...how do native elements behave here?

chrisjbrown:
> @chrisjbrown that's a good point about the order of events leading to a change in behavior...how do native elements behave here?

natively seems be the other way 😕 
so perhaps i've missed something here

daneah:
That makes sense I guess, and means that the native event does change the element's state but still allows it to be canceled (and therefore undone). Could we update the state first, but set it back if it gets canceled?

chrisjbrown:
> That makes sense I guess, and means that the native event does change the element's state but still allows it to be canceled (and therefore undone). Could we update the state first, but set it back if it gets canceled?

I was thinking the same. But event.target is a reference to the dom object and it becomes kinda unclear what's going on when the 'checked' value is true before the prevent default  and false after

Still looking for a better solution

chrisjbrown:
this latest change works for my needs. and i don't think would be a breaking change for others... 🤔 
I don't love it's kinda confusing... but not sure i have any better ideas

chrisjbrown:
> @chrisjbrown I think this makes sense to me; can you give an example of the expected usage on the consumer?

sure thing.
Consumers who do not `preventDefault()` will see no difference other than the addition of the `event.detail.target` 
Consumers who do `preventDefault()` will see that that as soon as `preventDefault()` is called the pharos-checkbox will not be `checked` (or vice versa depending on it's original state). 

Either way a new property will be available on the `change` Event in the form of `event.detail.target` which will refer to the original native checkbox should the consumer need to access it.

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -64,15 +64,23 @@ export class PharosCheckbox extends FormMixin(FormElement) {
   }
 
   public onChange(): void {
-    this.indeterminate = false;
     this.checked = this._checkbox.checked;
+    this.indeterminate = false;
```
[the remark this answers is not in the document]
chrisjbrown: 🤔 yeah i think we do.
just added that

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -64,15 +64,25 @@ export class PharosCheckbox extends FormMixin(FormElement) {
...
         composed: true,
+        detail: {
+          target: this._checkbox,
+        },
       })
     );
+
+    if (!notCancelled) {
+      this.indeterminate = originalIndeterminateState;
+      this.checked = !this._checkbox.checked;
```
daneah, line 84: Wonder if we want to update this one to use the same "original state reset" form as the indeterminate, for consistency

chrisjbrown: is this what you meant?
https://github.com/ithaka/pharos/pull/431/files#diff-18ffa349736aa11f0cc47dfdcb4e2129f167d12b920d4e7e82ead64a6734a524R67-R68

i do think that reads better

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -64,6 +64,9 @@ export class PharosCheckbox extends FormMixin(FormElement) {
   }
 
   public onChange(): void {
+    /**
+     * store the original indeterminate and checked states
```
daneah, line 68:
```suggestion
     * Store the original indeterminate and checked states
     * The native checkbox state has already changed, so invert it
```

daneah (approved):

PR #435, opened 2022-10-31
Storybook: Convert stories to CSF 3

### LINKED ISSUE #244: Convert stories to Component Story Format
**The problem**
To keep development snappy and utilize the [latest features Storybook provides](https://storybook.js.org/blog/storybook-on-demand-architecture/) we need to move away from the current `mdx` format used for stories.

**The solution**
Refactor stories to [CSF 3.0 format](https://storybook.js.org/blog/component-story-format-3-0/).

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Resolves #244 

> **Note**
>
> **Note to reviewers**: The fact that Chromatic reports on visual changes and isn't reporting on much here should be a strong signal of parity; the main reported differences were intentional ones due to improvements I made along the way, or were otherwise trivial ones. You can see the full change set [here](https://www.chromatic.com/pullrequest?appId=60919c26122bd50039b34644&number=435&view=changes).

Convert `Alert` stories from MDX to CSF.

Compare the [new page](https://60919c26122bd50039b34644-cycpwypwzr.chromatic.com/?path=/docs/components-alert--closable) to the [current page](https://pharos.jstor.org/storybooks/wc/?path=/docs/components-alert--info).

Notable improvements:

* Remove `source: { type: 'code' } }` from the React story configuration so that the **Code** option below examples shows copy/pastable code instead of some raw story code. This is how the web component stories already work.
* We get most all of the value of how the **Docs** tab used to work, with less code and configuration (ostensibly).
* The web component Storybook will now automatically show the component's JSDoc description at the top of its **Docs** page. I can't think of a way to make that work for the React component beyond doing dynamic code generation instead of a functional wrapper like we do now.
* Several inconsistencies between the web component and React component stories, such as missing stories or value differences, have been freshly aligned.
* Many components' stories have had the `defaultArgs` and the `argTypes` extracted to a `storyArgs` module so both the web component and the React component story can make use of them.

Some things to look out for:

* It's more cumbersome to limit which controls show up for a given story—it used to be just those specified in `argTypes`, but now it's all of them unless otherwise specified in a story's `include/exclude` parameters. Maybe this is fine; more stories end up being something you can fiddle with and link to for collaborative purposes, which is useful.
* It's no longer easy to sort stories in a very specific order. They're easiest to sort alphabetically (default), and anything aside from that requires coming up with some kind of scheme.
* `addon-storysource` has not yet caught up to CSF 3 and doesn't display any code in the **Story** panel when using the **Canvas**. The code block on the **Docs** tab does still work, however.
  * Related issue: storybookjs/storybook#17975

**Additional context**

The two story files look more similar than ever—the only difference is really that one uses e.g. the `<pharos-alert>` web component while the other uses e.g. the `<PharosAlert>` React component.


jialin-he (approved):

.storybook/docsPageConfig.jsx
```diff
@@ -0,0 +1,25 @@
...
+      <Title />
+      <Subtitle />
+      <Description />
+      {componentName && <GuidelineLink path={componentName} />}
+      <Primary />
+      <ArgsTable story={PRIMARY_STORY} />
+      <Stories />
+    </>
+  );
+};
```
daneah, line 25: This is a noteworthy file; I extracted the common ordering of items for the **Docs** tab of each component into a reusable piece. This was previously specified implicitly in each story using MDX, but now gets added in the `parameters.docs.page` configuration of the default export for each component.

.storybook/main.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  features: {
+    previewCsfV3: true,
+  },
```
daneah, line 4: This allows us to use Component Story Format (CSF) version 3 in the stories, and will become the default in Storybook v7.

satya-achanta-venkata (approved):

chrisjbrown (approved):

jialin-he (approved):

PR #438, opened 2022-11-02
ToggleButton: indicate toggle button state on underlying button elements

### LINKED ISSUE #430: Toggle button group not conveying state
**Expected behavior**
The [Toggle Button Group](https://pharos.jstor.org/components/toggle-button-group) conveys the active/inactive status of each control both visually and programmatically.

**Actual behavior**
These controls do not programmatically convey their current state and are only announced with name and button role, ex. "List, button; Grid, button". Screen reader users will not know which option is currently selected.

1. While using a screen reader, navigate to the List/Grid control
2. Navigate between controls
3. Observe announcement of Name and Role, but does not announce state.

**Screenshots or code**
Screen Shot 2022-10-20 at 11 26 12 AM

**Pharos version**
Production

**Your environment**
OS: macOS, Windows
Browser: Chrome, Firefox, Safari
Version: 97.0, 105.0.2, 16.0

**Additional information**
Developers must ensure that the active/inactive or selected state of a control is conveyed programmatically. Most semantic input elements have this feature built in, checkbox, radio button, etc. Recommendation would be to treat this control as radio buttons, keep the same visual styling, but change the inputs accordingly. If this is not possible, or for future custom controls, developers can use the aria-selected attribute to convey this information specifically for users of assistive technology.

daneah:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Resolves #430 

Because the accessibility object model (AOM) is still in active development and doesn't enjoy wide browser vendor adoption at present, we need to maintain the accessibility for assistive technologies in the underlying native elements rather than the custom element host containers.

For things like buttons, this often means adding a property that gets rendered onto the underlying native element as an `aria-` attribute. Here, I add a `pressed` property to Button so that the ToggleButton can set it appropriately, which gets rendered as `aria-pressed` on the underlying `<button>` element. I also add a `group-label` prop to ToggleButton that renders the `aria-label` for the full button group, useful for consumers to add the most contextualized description they can for each usage.

**Additional context**

The approach here is a generally desired one for optimal accessibility, and not unique to this particular component. Some of these may prove more difficult to manage, such as getting the right `aria-describedby` value down into the native element for any and every component that might need a tooltip. One approach could be for the tooltip to attach itself to the outermost shadow DOM element, but we'll need to take stock to see how future-proof that'd be.

### DIFF
--- a/packages/pharos/src/components/button/pharos-button.ts
+++ b/packages/pharos/src/components/button/pharos-button.ts
@@ -17,6 +17,8 @@ export type ButtonType = 'button' | 'submit' | 'reset';
 
 export type ButtonVariant = 'primary' | 'secondary' | 'subtle' | 'overlay';
 
+type PressedState = 'false' | 'true' | 'mixed' | 'undefined';
+
 const TYPES = ['button', 'submit', 'reset'];
 
 const VARIANTS = ['primary', 'secondary', 'subtle', 'overlay'];
@@ -136,6 +138,13 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
   @property({ type: String, reflect: true })
   public value?: string;
 
+  /**
+   * Indicates this button is a toggle button and whether it is pressed or not.
+   * @attr value
+   */
+  @property({ type: String, reflect: true })
+  public pressed: PressedState = 'undefined';
+
   @query('#button-element')
   private _button!: HTMLButtonElement | HTMLAnchorElement;
 
@@ -245,6 +254,7 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
             rel=${ifDefined(this.rel)}
             target=${ifDefined(this.target)}
             aria-label=${ifDefined(this.label)}
+            aria-pressed=${ifDefined(this.pressed)}
             @keyup=${this._handleKeyup}
           >
             ${this.buttonContent}
@@ -259,6 +269,7 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
             ?disabled=${this.disabled}
             type="${ifDefined(this.type)}"
             aria-label=${ifDefined(this.label)}
+            aria-pressed=${ifDefined(this.pressed)}
           >
             ${this.buttonContent}
           </button>
--- a/packages/pharos/src/components/button/pharos-button.ts
+++ b/packages/pharos/src/components/button/pharos-button.ts
@@ -17,7 +17,7 @@ export type ButtonType = 'button' | 'submit' | 'reset';
 
 export type ButtonVariant = 'primary' | 'secondary' | 'subtle' | 'overlay';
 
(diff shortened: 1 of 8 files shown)


packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -17,6 +17,8 @@ export type ButtonType = 'button' | 'submit' | 'reset';
 
 export type ButtonVariant = 'primary' | 'secondary' | 'subtle' | 'overlay';
 
+export type PressedState = 'false' | 'true' | 'mixed' | 'undefined';
```
jialin-he, line 20: Curious what state is a "mixed" state?

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -136,6 +138,13 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
   @property({ type: String, reflect: true })
   public value?: string;
 
+  /**
+   * Indicates this button is a toggle button and whether it is pressed or not.
+   * @attr value
+   */
+  @property({ type: String, reflect: true })
+  public pressed: PressedState = 'undefined';
```
chrisjbrown, line 146: should this be set to the string `undefined`?

chrisjbrown (approved):

sirrah-tam (commented):
I was able to validate this update and confirm the expected behavior with VoiceOver on MacOS in Firefox and Safari and on Windows with NVDA and Chrome/Firefox/Edge. 

However, this appears to have introduced a bug in Chrome (Version 97.0.4692.71) with VO where none of the toggle buttons have a calculated name and are just announced as "Toggle button" along with current state.

daneah:
@sirrah-tam Thanks for looking! Interestingly, I observed Chrome working with VO on my end 🤔

sirrah-tam:
Hmm, it appears that my installation of Chrome on my work laptop has been kept back by some org settings. Tested this on my personal laptop and can confirm its working as expected. Tested on Chrome v 107.0.5304.87.

sirrah-tam (approved):
Validated and working as expected. Screen reader users will perceive the group name upon entering the toggle button grouping and each control is announced as a toggle button and includes its current state as pressed/not pressed.

jialin-he (approved):

sirrah-tam:
@daneah - One side affect of the change we made to this is that all Pharos buttons now have the aria-pressed="undefined" attribute added by default. While this isn't necessarily problematic at first, it has shown a bit of inconsistency across some browser & AT combinations. While testing with VoiceOver on Safari, the presence of that attribute makes the button be parsed as a "toggle button" and is announced as such even though I don't believe that's the intention. Users will expect certain behavior if its announced as that type of button control, and could become confused if it does something different.

daneah:
@sirrah-tam can you please open a new issue?

PR #470, opened 2023-01-09
Combobox: add loose-match option for matching query

NazimHAli:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Provides the option to loosely match user input values to the text options. By default there's an exact match and this excludes cases when a query or text has accents or different casing.

Added method to normalize strings using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#parameters

**Additional context**
Users couldn't find their institutions when searching on https://www.jstor.org/site/collection-list/ because the official spelling contains an accent.

### DIFF
--- a/packages/pharos/src/components/combobox/pharos-combobox.ts
+++ b/packages/pharos/src/components/combobox/pharos-combobox.ts
@@ -68,6 +68,13 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement)) 
   public searchMode = false;
 
   /**
+   * Use loose matching when comparing input value to options.
+   * @attr looseMatch
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'loose-match' })
+  public looseMatch = false;
+
+  /**
    * The list of options available in the combobox dropdown list
    * @readonly
    */
@@ -166,9 +173,15 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement)) 
 
   private _renderList(): TemplateResult | typeof nothing {
     if (this.open) {
-      const regex = new RegExp(this._query, 'gi');
+      const queryToCompare = this.looseMatch ? this._normalizeString(this._query) : this._query;
+      const regex = new RegExp(queryToCompare, 'gi');
       const matchingOptions = this.options.filter((child) => {
-        return this.searchMode ? child : child.text.match(regex);
+        if (this.searchMode) {
+          return child;
+        }
+
+        const childText = this.looseMatch ? this._normalizeString(child.text) : child.text;
+        return childText.match(regex);
       });
       this._noResults = matchingOptions.length === 0;
 
@@ -433,6 +446,13 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement)) 
 
   private _handleInputClick(): void {
     this.open = !this.open;
+  }
+
+  private _normalizeString(textString: string) {
+    return textString
+      .normalize('NFKD')
+      .replace(/[\u0300-\u036f]/g, '')
+      .toLowerCase();
   }
 
   protected override render(): TemplateResult {
--- a/packages/pharos/src/components/combobox/pharos-combobox.ts
+++ b/packages/pharos/src/components/combobox/pharos-combobox.ts
@@ -448,6 +448,15 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement)) 
     this.open = !this.open;
   }
 
+  /**
+   * Normalizes a string in the following order:
+   *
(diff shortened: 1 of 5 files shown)


packages/pharos/src/components/combobox/pharos-combobox.ts
```diff
@@ -435,6 +448,13 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement))
     this.open = !this.open;
   }
 
+  private _normalizeString(textString: string) {
+    return textString
+      .normalize('NFKD')
+      .replace(/[\u0300-\u036f]/g, '')
+      .toLowerCase();
+  }
```
NazimHAli, line 456: There's also the method `localeCompare` that filters accents - testing it now and might removed the need for this method.

packages/pharos/src/components/combobox/pharos-combobox.ts
```diff
@@ -435,6 +448,13 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement))
     this.open = !this.open;
   }
 
+  private _normalizeString(textString: string) {
+    return textString
+      .normalize('NFKD')
+      .replace(/[\u0300-\u036f]/g, '')
```
daneah, line 454: Might you leave a comment here explaining this? I don't know offhand what it does, myself! Is that a range of unprintable unicode characters?

daneah (approved):
This looks really nice, thank you for adding it! I have one suggestion for future maintainability

packages/pharos/src/components/combobox/pharos-combobox.ts
NazimHAli: `localeCompare` is for a different use case, not partial matching.

packages/pharos/src/components/combobox/pharos-combobox.ts
NazimHAli: It's a range of all possible diacritics - after text is normalized, than we can remove the characters and compare the text.

packages/pharos/src/components/combobox/pharos-combobox.ts
NazimHAli: Testing variations of this method, there are `search, sensitivity` options https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#options

NazimHAli: The search option won't do partial matching either.

```js
"nèat".localeCompare("neat now", { usage: "search", ignorePunctuation: true, sensitivity: "base" })
```

packages/pharos/src/components/combobox/pharos-combobox.ts
daneah: Please leave this as a code comment, if you don't mind!

brentswisher (approved):
Nice work! I had struggled a little with this in the past (In Oracle, not JS) so I was curious and this looks really good. I tried this with some cyrillic characters that caused issues then and it handled them well. 

I did try some things which are a step further than removing the accent mark, like seeing if the Germanic ß matches "ss" and that didn't appear to work - but I don't think that is necessary here. 

Just mentioning this because I find it interesting, I don't think it is worth adding a package dependency to fix.  This article goes into detail about it, but essentially `normalize` takes off the extra unicode diacritic and leaves the base letter, and there are some weird edge cases like `ß` and `ł` where that doesn't work: https://javascript.plainenglish.io/not-so-obvious-removal-of-diacritics-in-javascript-explained-and-done-right-52f4aeb3c85

NazimHAli:
> Nice work! I had struggled a little with this in the past (In Oracle, not JS) so I was curious and this looks really good. I tried this with some cyrillic characters that caused issues then and it handled them well.
> 
> I did try some things which are a step further than removing the accent mark, like seeing if the Germanic ß matches "ss" and that didn't appear to work - but I don't think that is necessary here.
> 
> Just mentioning this because I find it interesting, I don't think it is worth adding a package dependency to fix. This article goes into detail about it, but essentially `normalize` takes off the extra unicode diacritic and leaves the base letter, and there are some weird edge cases like `ß` and `ł` where that doesn't work: https://javascript.plainenglish.io/not-so-obvious-removal-of-diacritics-in-javascript-explained-and-done-right-52f4aeb3c85

Thank you for sharing! As we increase the amount of content, I think we'll get to a point where we will need to address the cases you mentioned. We already have some collections in Arabic, Farsi, Hindi/Urdu/Benghali, (I don't know offhand if we have content in Germanic, but assume we do because of European collections) that need to be searchable.

PR #472, opened 2023-01-13
Infra: Bump TypeScript, add main exports

### LINKED ISSUE #212: Add exports field to package.json
**The problem**
Webpack 5 module federation dependency sharing works best with imports from a package's root entry file. Currently to share Pharos, folks would need to list the individual components they'd like to share as their imports go deep into the package:

```js
import '@ithaka/pharos/lib/components/modal/pharos-modal';
```

**The solution**
Address #3 and then add an [exports](https://nodejs.org/api/packages.html#packages_exports) field to the package.json to define the main entry point and encapsulate the package to prevent other entry points.

```json
"exports": {
    ".": "./lib/index.js",
    "./lib": "./lib/index.js",
    "./lib/index": "./lib/index.js",
    "./lib/index.js": "./lib/index.js",
    "./package.json": "./package.json",
    "./custom-elements.json": "./custom-elements.json"
  }
```

The first entry would allow consumers to import component definitions in this way:

```js 
import { PharosButton } from '@ithaka/pharos';
```

Thus they would only need to add `@ithaka/pharos` as a shared dependency in their MFE configuration.

**Alternatives considered**
N/A.

**Additional information**
This will likely be a breaking change as only the entry points defined in the field would be allowed.

chrisjbrown:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Resolves #212

[...]


packages/pharos-site/static/styles/global.scss
```diff
@@ -1,5 +1,6 @@
 @use '@ithaka/pharos/lib/styles/pharos';
-@use '@ithaka/pharos/lib/utils/scss/mixins';
```
chrisjbrown, line 2: this should not be commented out
but scss-loader isn't able to find this file and i'm not sure why 😕

packages/pharos/src/pages/search/search.scss
```diff
@@ -1,4 +1,4 @@
-@use '@ithaka/pharos/lib/utils/scss/mixins';
+// @use '@ithaka/pharos/lib/utils/scss/mixins';
```
chrisjbrown, line 1: this should not be commented out
but scss-loader isn't able to find this file and i'm not sure why 😕

packages/pharos-site/static/styles/global.scss
chrisjbrown: fixed
needed to include the "_" in the export

chrisjbrown:
🤔  
failing to build the pharos-site `yarn site:build`
```
 ERROR #98124  WEBPACK

Generating JavaScript bundles failed

undefined

If you're trying to use a package make sure that 'undefined' is installed. If you're
trying to use a local file make sure that the path is correct.

File: src/components/Sidenav.tsx:1:470
```

packages/pharos/package.json
```diff
@@ -84,7 +84,19 @@
...
-    "typescript": "^4.5.2"
+    "typescript": "^4.9.4"
   },
-  "customElements": "custom-elements.json"
+  "customElements": "custom-elements.json",
+  "exports": {
+    ".": "./lib/index.js",
+    "./lib": "./lib/index.js",
+    "./lib/index": "./lib/index.js",
+    "./lib/index.js": "./lib/index.js",
```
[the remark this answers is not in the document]
chrisjbrown: thanks @Niznikr 
that's one step closer.
now seeing some errors for layout.tsx as well as errors around mdx files

```
 ERROR #98124  WEBPACK

Generating JavaScript bundles failed

undefined

If you're trying to use a package make sure that 'undefined' is installed. If you're
trying to use a local file make sure that the path is correct.

File: src/components/layout.tsx:1:231
```

```
 ERROR #98124  WEBPACK

Generating JavaScript bundles failed

undefined

If you're trying to use a package make sure that 'undefined' is installed. If you're
trying to use a local file make sure that the path is correct.

File: src/pages/design-tokens/alias-colors.mdx:10:0
```

packages/pharos/package.json
```diff
@@ -84,7 +84,20 @@
...
+    "./lib/react-components": "./lib/react-components/index.js",
+    "./lib/styles/pharos": "./lib/styles/pharos.scss",
+    "./lib/styles/pharos.scss": "./lib/styles/pharos.scss",
+    "./lib/utils/scss/mixins": "./lib/utils/scss/_mixins.scss",
```
NazimHAli, line 98: `yarn build:site` is failing because when it tries to build, it expects other files to be exported. Using patterns would be useful here if really do need to import all of these files https://nodejs.org/api/packages.html#subpath-patterns

Running site build locally passed with warnings instead of errors:

```diff
--- a/packages/pharos/package.json
+++ b/packages/pharos/package.json
@@ -95,7 +95,22 @@
     "./lib/react-components": "./lib/react-components/index.js",
     "./lib/styles/pharos": "./lib/styles/pharos.scss",
     "./lib/styles/pharos.scss": "./lib/styles/pharos.scss",
+    "./lib/styles/typography": "./lib/styles/pharos.scss",
+    "./lib/styles/typography.scss": "./lib/styles/typography.scss",
+    "./lib/styles/fonts.css": "./lib/styles/fonts.css",
+    "./lib/styles/variables": "./lib/styles/variables.css",
+    "./lib/styles/variables.css": "./lib/styles/variables.css",
+    "./lib/styles/variables.js": "./lib/styles/variables.js",
+    "./lib/styles/pharos-alert.css": "./lib/styles/pharos-alert.css",
+    "./lib/styles/pharos-footer.css": "./lib/styles/pharos-footer.css",
+    "./lib/styles/pharos-modal.css": "./lib/styles/pharos-modal.css",
+    "./lib/styles/tokens": "./lib/styles/tokens.js",
+    "./lib/styles/tokens.js": "./lib/styles/tokens.js",
     "./lib/utils/scss/mixins": "./lib/utils/scss/_mixins.scss",
+    "./lib/utils/registerComponents": "./lib/utils/registerComponents.js",
+    "./lib/utils/registerComponents.js": "./lib/utils/registerComponents.js",
+    "./lib/utils/PharosContext": "./lib/utils/PharosContext.js",
+    "./lib/utils/PharosContext.js": "./lib/utils/PharosContext.js",
     "./lib/utils/scss/mixins.scss": "./lib/utils/scss/_mixins.scss",
     "./package.json": "./package.json",
     "./custom-elements.json": "./custom-elements.json"
```

.storybook/theme.js
```diff
@@ -6,7 +6,7 @@ import {
   PharosColorMarbleGray50,
   PharosColorMarbleGray20,
   PharosFontFamilyBody,
-} from '../packages/pharos/lib/styles/variables.js';
+} from '../packages/pharos/lib/styles/variables';
 import '../packages/pharos/lib/styles/fonts.css';
```
chrisjbrown, line 10: tried to standardize the exports.
what i have so far is `.js` should always drop the extension and anything else should always include the extension

packages/pharos/package.json
```diff
@@ -84,7 +84,28 @@
...
+    "./lib/react-components": "./lib/react-components/index.js",
+    "./lib/styles/pharos": "./lib/styles/pharos.scss",
+    "./lib/styles/typography": "./lib/styles/pharos.scss",
+    "./lib/styles/typography.scss": "./lib/styles/typography.scss",
+    "./lib/styles/fonts.css": "./lib/styles/fonts.css",
+    "./lib/styles/variables.css": "./lib/styles/variables.css",
+    "./lib/styles/variables": "./lib/styles/variables.js",
+    "./lib/styles/pharos-alert.css": "./lib/styles/pharos-alert.css",
+    "./lib/styles/pharos-footer.css": "./lib/styles/pharos-footer.css",
+    "./lib/styles/pharos-modal.css": "./lib/styles/pharos-modal.css",
```
daneah, line 103: I'd bump @NazimHAli's suggestion about using patterns for these, as not remembering to do this will likely result in some patch fixes in the future. Unless there was a reason that didn't work?

chrisjbrown: so this seems to work
```
    "./lib/styles/*": "./lib/styles/*.js",
    "./lib/styles/*.css": "./lib/styles/*.css",
    "./lib/styles/*.scss": "./lib/styles/*.scss",
    "./lib/styles/pharos": "./lib/styles/pharos.scss",
  ```
  i needed to export styles/pharos.scss separately because storybook was complaining it couldn't find it with `@use '@ithaka/pharos/lib/styles/pharos';`

**Review by daneah — APPROVED** [review]

File: packages/pharos/package.json [hunk 7]
```diff
@@ -84,7 +84,22 @@
     "sinon": "^12.0.1",
     "style-dictionary": "^3.0.1",
     "ts-lit-plugin": "^1.2.1",
-    "typescript": "^4.5.2"
+    "typescript": "^4.9.4"
   },
-  "customElements": "custom-elements.json"
+  "customElements": "custom-elements.json",
+  "exports": {
```
Niznikr, line 90: Think you'll want to add a changeset and designate it as [a breaking change](https://nodejs.org/api/packages.html#package-entry-points) since individual component imports will be invalid.

Niznikr (approved):

PR #475, opened 2023-01-19
Toast: Update Toaster to use internally consistent scoping

### LINKED ISSUE #474: PharosToaster doesn't register PharosToast in the scope of its shadow root
**Expected behavior**
PharosToaster registers PharosToast in the scope of its shadow root, so that it can use the PharosToast component without needing the parent application to register PharosToast.

**Actual behavior**
PharosToast assumes that the PharosToast component has already been registered by the parent application using the same prefix as itself. [See these lines of the PharosToaster source code](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/toast/pharos-toaster.ts#L60-L61). If the parent application doesn't register PharosToast, then the PharosToast components PharosToaster creates will not have styling or logic since the browser doesn't recognize their custom HTML element tags.

**Additional information**
PharosToaster can use [@lit-labs/scoped-registry-mixin](https://github.com/lit/lit/tree/main/packages/labs/scoped-registry-mixin) to register PharosToast in the scope of its shadow root. PharosToast already uses this method to register PharosIcon and PharosToastButton. [See these lines of the PharosToast source code](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/toast/pharos-toast.ts#L41-L45).

daneah:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Fixes #474 

Updates `PharosToaster` to use `ScopedRegistryMixin` and replace (most of) the DOM manipulation it was doing with declarative HTML rendering.

**Additional context**

This is a breaking change because it changes the `detail` value for the `pharos-toast-close` event. Because this event was externally visible, this is technically an API change. I don't, however, believe any internal consumers were listening for this event. Because we're gearing up for v13 and this isn't otherwise too pressing, let's just wait and include it there.

### DIFF
--- a/packages/pharos/src/components/toast/pharos-toaster.ts
+++ b/packages/pharos/src/components/toast/pharos-toaster.ts
@@ -148,7 +148,7 @@ export class PharosToaster extends ScopedRegistryMixin(PharosElement) {
 
   private _renderToast(toast: ToastDetail): TemplateResult {
     // This is used to properly render any Pharos components present in the supplied content.
-    // unsafeHtml will render content in general, but does not appear to render components.
+    // unsafeHTML will render content in general, but does not appear to render components.
     const toastContentElement = document.createElement('div');
     toastContentElement.innerHTML = toast.content;
 
--- a/packages/pharos/src/components/toast/pharos-toaster.ts
+++ b/packages/pharos/src/components/toast/pharos-toaster.ts
@@ -1,32 +1,83 @@
 import { PharosElement } from '../base/pharos-element';
 import { html } from 'lit';
 import type { TemplateResult, CSSResultArray } from 'lit';
+import { state } from 'lit/decorators.js';
+import { repeat } from 'lit/directives/repeat.js';
 import { toasterStyles } from './pharos-toaster.css';
 
-import type { PharosToast } from './pharos-toast';
+import type { ToastStatus } from './pharos-toast';
+import { PharosToast } from './pharos-toast';
 import { DEFAULT_STATUS, DEFAULT_INDEFINITE } from './pharos-toast';
 
 import { v4 as uuidv4 } from 'uuid';
+import ScopedRegistryMixin from '../../utils/mixins/scoped-registry';
 
 /**
  * pharos-toast-open event.
  *
- * @tag pharos-toaster
- *
  * @event pharos-toast-open
- * @type {object}
- * @property {ToastStatus} status - The status for the toast.
- * @property {string} content - The content to slot into the toast.
+ * @type {ToastCreateDetail}
  */
+
+/**
+ * pharos-toast-update event.
+ *
+ * @event pharos-toast-update
+ * @type {ToastUpdateDetail}
+ */
+
+/**
+ * pharos-toast-close event.
+ *
+ * @event pharos-toast-close
+ * @type {ToastCloseDetail}
+ */
+
+type ToastID = string;
+type ToastContent = string;
+type ToastIndefinite = boolean;
+
+interface ToastDetail {
+  id: ToastID;
+  content: ToastContent;
+  status: ToastStatus;
+  indefinite: ToastIndefinite;
+}
+
+interface ToastCreateDetail {
+  id?: ToastID;
+  content: ToastContent;
+  status?: ToastStatus;
+  indefinite?: ToastIndefinite;
+}
+
+interface ToastUpdateDetail {
+  id: ToastID;
+  content?: ToastContent;
+  status?: ToastStatus;
+}
+
+interface ToastCloseDetail {
+  id: ToastID;
+}
 
 /**
  * Pharos toaster component.
  *
- * @slot - Contains the toasts (the default slot).
+ * @tag pharos-toaster
  *
- * @listens pharos-toast-open
+ * @listens pharos-toast-open - Use this to create new toasts
+ * @listens pharos-toast-update - Use this to update an existing toast
+ * @listens pharos-toast-close - Use this to close an existing toast
  */
-export class PharosToaster extends PharosElement {
+export class PharosToaster extends ScopedRegistryMixin(PharosElement) {
+  static elementDefinitions = {
+    'pharos-toast': PharosToast,
+  };
+
+  @state()
+  private _toasts: ToastDetail[] = [];
+
   constructor() {
     super();
     this._openToast = this._openToast.bind(this);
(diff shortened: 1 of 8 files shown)


packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -1,32 +1,83 @@
...
+
+interface ToastUpdateDetail {
+  id: ToastID;
+  content?: ToastContent;
+  status?: ToastStatus;
+}
+
+interface ToastCloseDetail {
+  id: ToastID;
+}
```
daneah, line 62: Open to toning this down or suggestions on where to move these—we don't have an established pattern. I could see there being a good case for something like `src/components/<comp>/pharos-<comp>.types.ts`.

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -1,32 +1,83 @@
...
+ * @event pharos-toast-update
+ * @type {ToastUpdateDetail}
+ */
+
+/**
+ * pharos-toast-close event.
+ *
+ * @event pharos-toast-close
+ * @type {ToastCloseDetail}
  */
```
daneah, line 34: These were missing before.

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -1,32 +1,83 @@
...
+import { PharosToast } from './pharos-toast';
 import { DEFAULT_STATUS, DEFAULT_INDEFINITE } from './pharos-toast';
 
 import { v4 as uuidv4 } from 'uuid';
+import ScopedRegistryMixin from '../../utils/mixins/scoped-registry';
 
 /**
  * pharos-toast-open event.
  *
- * @tag pharos-toaster
```
daneah, line 14: I think this was errantly placed here; I moved it to the component's JSDoc block below.

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -52,45 +103,72 @@ export class PharosToaster extends PharosElement {
...
+    const toastContentElement = document.createElement('div');
+    toastContentElement.innerHTML = toast.content;
+
+    return html`<pharos-toast
+      id="${toast.id}"
+      status="${toast.status}"
+      ?indefinite="${toast.indefinite}"
+    >
+      ${toastContentElement}
+    </pharos-toast>`;
```
daneah: @Niznikr curious if you might understand more about why this is and if I've missed some trick here.

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -52,45 +103,72 @@ export class PharosToaster extends PharosElement {
...
```
daneah, line 161: @Niznikr curious if you might understand more about why this is and if I've missed some trick here—the following did not work, failing to render e.g. a `pharos-link` present in `toast.content`:

```js
    return html`<pharos-toast
      id="${toast.id}"
      status="${toast.status}"
      ?indefinite="${toast.indefinite}"
    >
      ${unsafeHTML(toast.content)}
    </pharos-toast>`;
```

packages/pharos/src/components/toast/pharos-toaster.ts
eslawski: Personally like to keep types close to where they are used. We could consider moving them out into their own module when we need to use them elsewhere or wish to expose them to the consumer?

eslawski (approved):

packages/pharos/src/components/toast/pharos-toaster.ts
Niznikr: Not certain tbh. Believe I ran into this the first time around. The [Lit Discord](https://discord.com/invite/buildWithLit) might be helpful.

daneah: Interestingly this doesn't happen in [the Lit playground](https://lit.dev/playground/?#gist=46422f0c6361d69614f57d0e66cef633).

chrisjbrown: is this because pharos-link isn't being registered anywhere in the story?

if I add `PharosLink` as a "local" definition in the toast. it does render correctly
```
export class PharosToaster extends ScopedRegistryMixin(PharosElement) {
  static elementDefinitions = {
    'pharos-toast': PharosToast,
    'pharos-link': PharosLink,
  };
```

** just did this to test if it would work, not suggesting this

daneah: > is this because pharos-link isn't being registered anywhere in the story?

The components are all registered and available to all stories by nature of [this module](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/test/initComponents.ts) and [its import](https://github.com/ithaka/pharos/blob/develop/.storybook/main/preview.js#L9).

> if I add PharosLink as a "local" definition in the PharosToaster. it does render correctly

To be clear, it already does render correctly currently. The question is whether there's a way to avoid the DOM manipulation here while maintaining the functionality.

The crux being that the supplied content comes from the consumer, and the toast _cannot predict_ what that content will be, and therefore does not want to be in the business of e.g. registering all possible components that could be in that content.

chrisjbrown: in my testing i added this
```
export class PharosToaster extends ScopedRegistryMixin(PharosElement) {
  static elementDefinitions = {
    'pharos-toast': PharosToast,
    'pharos-link': PharosLink,
  };
```

as well as changed this back to 
```
return html`<pharos-toast
      id="${toast.id}"
      status="${toast.status}"
      ?indefinite="${toast.indefinite}"
    >
      ${unsafeHTML(toast.content)}
    </pharos-toast>`;
```
```

daneah: I have been assuming this is due to the supplied content being put into the `Toast`'s light DOM, but now that you prompt me to think more about it the `Toast` is inside the `Toaster`'s shadow DOM so I am not sure exactly what I should be expecting. The scope seems like it would indeed be the `Toaster`'s registry.

daneah: (Updated in the [Discord discussion](https://discord.com/channels/1012791295170859069/1065652155257737286) as well) @chrisjbrown and I tracked this down to the scoped registry behavior, which we now understand does not have inheritance—you either get the global registry without isolation, or you get your total isolation with only explicit components allowed. The DOM manipulation style is what allows us to straddle the two.

PR #478, opened 2023-01-23
Sidenav: remove sidenav button

### LINKED ISSUE #417: Sidenav: Change to consumer-defined closing behavior
**The problem**
I am frustrated with how the sidenav component tries to control its own state when not knowing the context in which a user will be using it. It gives the option to control its visibility through the _slide_ prop, but does not emit an event for when the _close button_ within the sidenav is clicked. This renders any manual control of state for the sidenav useless since there will always be a case where a user can close the sidenav without the containing app knowing. 

The benefits of the sidenav are that it organizes hierarchical menu items in a easy to use and good looking fashion. It gives users the option to separate categories into divided sections. The sidenav for those reasons already offers enough use to the user. I don't believe that trying to control its own state should be a function of the sidenav, but rather be deferred to the user's discretion. Function such as what the PharosSidenavButton provide and how the PharosSidenav automatically shows and hide based on device resolution may save time in specific contexts, but automatically removes control from the user and constrains the component in unnecessary ways.

**The solution**
Get rid of the PharosSidenavButton component. Eliminate any automatic toggling of sidenav state. Emit an event for when the _close button_ is clicked. This defers all control to the user (which is really easy, it takes no time to create a prop and event handler) and prevents any limitation in use.

**Alternatives considered**
At a minimum, add an event for when the _close button_ is clicked.

**Additional information**
I think it may be worthwhile to explain my specific use case where i ran into this issue...

While implementing the mobile redesign of the site navigation, I needed to use the sidenav as the new means of navigation. The site navigation would be collapsed into a hamburger button (PharosSidenavButton) and when pressed, would toggle the menu (PharosSidenav). However, when the sidenav is toggled I also need to toggle a backdrop behind it (to prevent interaction with components behind the sidenav as well as for aesthetics). Because I cannot invoke a callback for when the sidenav _close button_ is clicked, I cannot control both the sidenav and backdrop in tandem. Furthermore, adding a callback for the click event on the PharosSidenavButton seems to break its default functionality of toggling the sidenav state.

chrisjbrown:
This change removes the PharosSidenavButton and automatic sliding behavior of the sidenav. Before, the PharosSidenavButton and the PharosSidenav controlled its visibility automatically, making it difficult for consumers to control its state.

Now, users can control... 
- the visibility of the sidenav via the `open` property
- the visibility of the close button via the `hasCloseButton` property
- assign the `data-sidenav-id` attribute on trigger elements to open the sidenav when clicked

- [x] Adds a new feature
- [ ] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

Resolves #417

Removes `PharosSidenavButton` and it's references. Previously the button was being used to maintain the state of the `SideNav`. This change moves that responsibility to the consumer where it can pass `slide` = [true/false] in order to control the visibility of the `PharosSideNav`.


chrisjbrown:
🤔  not sure pharos-site is failing to build...
it's as if it's trying to build with pharos version 13 event though it's package.json lists 12.6.0

chrisjbrown:
I think this one's good to go

packages/pharos/src/components/sidenav/pharos-sidenav.ts
```diff
@@ -41,51 +46,73 @@ export class PharosSidenav extends ScopedRegistryMixin(FocusMixin(SideElement))
...
-    const details = {
-      bubbles: true,
-      composed: true,
-      detail: this,
-    };
-    this.dispatchEvent(new CustomEvent('pharos-sidenav-close', details));
+  private _addTriggerListeners(): void {
+    const id = this.getAttribute('id');
+    this._triggers = document.querySelectorAll(`[data-sidenav-id="${id}"]`);
+    console.log(this._triggers);
```
daneah, line 74:
```suggestion
```

daneah (commented):
We may want to replace what's currently here _removing_ the sidenav button on the Pharos site and stories with the new way of using a regular ol' button to trigger it opening. This might be separable from releasing the next version of the `@ithaka/pharos` package.

daneah (approved):

PR #479, opened 2023-01-24
DropdownMenu: add nav category to replace links

### LINKED ISSUE #323: DropdownMenuNavLink: Trying to activate dropdown navigates to link location on some mobile browsers/devices
**Expected behavior**

On mobile devices, "hovering" isn't a thing, so the proxy is a single tap to activate the hover behavior.

**Actual behavior**

On Android devices, a single tap on a DropdownMenuNavLink is instead clicking the link, causing the browser to navigate to the link's location instead of opening the dropdown. This is **not** the case on iOS.

1. Navigate to https://www.jstor.org
2. Attempt to tap the **Search** menu to open the options
3. Observe that you are immediately taken to the Advanced Search page instead

**Screenshots or code**
If applicable, add screenshots or code to help explain the issue.

**Pharos version**
At least v11+ though likely since the component was created

**Your environment**

- OS: Android
- Browser: (at least) Chrome, Firefox
- Version: Not sure

**Additional information**

First reported by @chrisjbrown 

This may have to do with the `aria-haspopup` attribute being on the custom element that triggers the dropdown, instead of the raw `<a>` element inside its shadow DOM. When the tap happens, the browser may not realize it should trigger the dropdown behavior instead of the link behavior as a result.

The setup of the `aria-haspopup` is all done from `PharosDropdownMenu`, and `PharosDropdownMenuNavLink` inherits from `PharosLink`, so it was difficult to quickly check what would happen if I just e.g. moved the `aria-haspopup` attribute into `PharosDropdownMenuNavLink`'s `<a>` element.

### LINKED ISSUE #354: DropdownMenuNav: Doesn't drop down when using a screen reader
**Expected behavior**
Dropdown menus should open without navigating to a new page

**Actual behavior**
Drop down menus in the main navigation don’t drop down. (Search, Browse, Tools). The only way to access these options using a screen reader is to click too open them in a new page - this causes confusion and extra work

1. (using JAWS) Go to JSTOR
2. tab to Search or Browse or Tools in the main nav
3. Press Enter or Space to open the menu. The site loads the linked page instead of opening the navigation. It doesn't open if I press the down arrow
4. On Chrome+VoiceOver I also can't consistently get the dropdown to work properly when i hit the spacebar or the down arrow. it tells me it's a visited link instead of "dropdown name, collapsed, menu item"

**Screenshots or code**
If applicable, add screenshots or code to help explain the issue.

**Pharos version**

Any

**Your environment**

- OS: Windows, mac
- Browser: Chrome

chrisjbrown:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

Possibly resolves #323 and possibly resolves #354

Caveat: This relies on consumers to not use `PharosDropdownMenuNavLink` but instead use `PharosDropdownMenuNavHeading` or anything else really.

Not sure if we can or would like to restrict what elements could be used in that slot


daneah (commented):
@chrisjbrown this looks like it could have legs to me; I think we do want to look further into the constraints you call out because ideally the developer will come away only having ~one right way to construct nav items (whether links or dropdowns). One way to do this might be to make the slot in the new component a named slot, so that only explicitly giving it the right name will properly render the dropdown content.

Aside: I think the use of the word `Heading` here serves to confuse when compared to actual headings that result in e.g. `<h1>` being rendered. You might consider `DropdownMenuNavCategory` or `DropdownMenuNavLabel` or some such.

@fswlee might you or other design folks have additional ideas? The purpose being to distinguish the pattern of "a thing that shows up and when hovered over shows a dropdown of links" (kind of like one of the two duties `DropdownMenuNavLink` currently serves) from the pattern of "I am a link directly in the nav that you can click on" (the think `DropdownMenuNavLink` will be going forward).

fswlee:
I will check with the design folks @daneah -- stay tuned

fswlee:
@daneah I thought the nav heading has a carat that shows when something is a dropdown?

chrisjbrown:
> @chrisjbrown this looks like it could have legs to me; I think we do want to look further into the constraints you call out because ideally the developer will come away only having ~one right way to construct nav items (whether links or dropdowns). One way to do this might be to make the slot in the new component a named slot, so that only explicitly giving it the right name will properly render the dropdown content.
> 
> Aside: I think the use of the word `Heading` here serves to confuse when compared to actual headings that result in e.g. `<h1>` being rendered. You might consider `DropdownMenuNavCategory` or `DropdownMenuNavLabel` or some such.
> 
> @fswlee might you or other design folks have additional ideas? The purpose being to distinguish the pattern of "a thing that shows up and when hovered over shows a dropdown of links" (kind of like one of the two duties `DropdownMenuNavLink` currently serves) from the pattern of "I am a link directly in the nav that you can click on" (the think `DropdownMenuNavLink` will be going forward).

added the slot name (i think correctly)
still not terribly restrictive. but i don't really see a way to restrict slots to certain types 🤔

jialin-he (approved):

chrisjbrown:
able to solve the AT click issue by using the private `hasHover` variable here
45360d2b71fe8b074f61b57d1205e9f13d5be448

though this exposed another issue where menu items are not being read correctly when navigating too them.
we suspect this may be caused by improper use of the `role` attribute. we're going to explore why that may be and if we can correct that in this PR
https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/

sirrah-tam:
> able to solve the AT click issue by using the private `hasHover` variable here [45360d2](https://github.com/ithaka/pharos/commit/45360d2b71fe8b074f61b57d1205e9f13d5be448)
> 
> though this exposed another issue where menu items are not being read correctly when navigating too them. we suspect this may be caused by improper use of the `role` attribute. we're going to explore why that may be and if we can correct that in this PR https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/

The issue exposed by this fix appears only to affect Chrome on both macOS and Windows. Both Safari and Firefox appear to work while using VoiceOver, announcing the menu item name as expected. While using Chrome, it appears that VoiceOver, JAWS, or NVDA are not able to properly calculated the menu items name.

It's not clear but the Dropdown Menu component works as expected, however the Dropdown Menu Nav does not. There may be an issue with how the Chrome browser is building the accessibility tree, which may make this issue difficult to fix.

sirrah-tam:
Based on some further testing and research, this appears to be an issue with how Chrome is parsing the specific menu option name and I will be submitting an issue to them directly.

With that said, I believe the fix made with this change is a significant enough improvement over the current navigation menu that I recommending moving forward with it and noting the potential issues some users may have.

jialin-he (approved):

sirrah-tam (approved):

PR #481, opened 2023-01-25
Site: fix site build

chrisjbrown:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

While making changes to pharos to update to typescript version `4.9.4` in #472 I also updated the `pharos-site/` directory to use the new `exports` from pharos 13. However, pharos-site is still on pharos 12, and 13 is not released yet.

I should have waited till pharos 13 is released, so that `pharos-site` can be updated accordingly with the available package.


chrisjbrown:
@daneah this is what i was mentioning earlier. reverting these `pharos-site` changes does seem to fix it though.

compared to other PR's that don't have this
https://github.com/ithaka/pharos/pull/479#issuecomment-1403750245

daneah:
@chrisjbrown makes sense. This is an interesting case caused by using published package versions in workspaces, rather than using the local dependency (which Yarn workspaces allows for). The ramifications of the latter being that any change to a package may necessitate a release of all dependent packages at the same time. The ramifications of **not** doing this are, right now, that we will need to remember to update the Pharos site sometime after v13 is published.

daneah (approved):

Issue #485, opened 2023-01-27
Focusing on a heading nested under a link doesn't render a blue outline in Chrome


NazimHAli:
**Expected behavior**
A blue outline on focus. Works correctly in Firefox:

<img width="287" alt="Screen Shot 2023-01-27 at 5 52 00 PM" src="https://user-images.githubusercontent.com/26750288/215221114-cc9806df-4859-43f4-9b28-60c664673754.png">

**Actual behavior**
There's no blue outline on focus

**Steps to reproduce the issue**

Reproducible in Chrome, but not on Firefox:

1. In Chrome, visit https://pharos.jstor.org/storybooks/wc/iframe.html?id=components-link--visited-link-heading&viewMode=story
2. Press tab until it focuses on the link
3. There's no blue outline on focus

**Pharos version**
12.17.0

**Your environment**
- Browser: Chrome
- Version: 109.0.5414.119

NazimHAli:
> did some digging and came up empty 😅 same issue it seems in Safari.
> 
> I do see the correct styles being added to the #link-element when programmatically selecting it. but for whatever reason it's not shown 😕

I looked into it and found the issue to be not rendering the layout and need to update the `display`. I will open a PR once I'm able to.

NazimHAli:
Additional findings from research/experimentation for future reference:
- From a styling perspective, it would be ideal to be able to style nested slots
- Currently we're not able to because it depends on open issues https://github.com/WICG/webcomponents/issues/936 and https://github.com/w3c/csswg-drafts/issues/6867

daneah:
We should consider how this is impacted as we think about using `:focus-visible` (#175)

PR #491, opened 2023-02-02
Link: display blue outline on focus

NazimHAli:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Resolves accessibility/display issue https://github.com/ithaka/pharos/issues/485

Blue outline not rendered in Chrome when focusing on a link with a nested heading.

Sets the `display` value to `inline-block` so that it renders the layout and elements dimensions (margin/padding) are applied properly.

### DIFF
--- a/.changeset/plenty-bobcats-know.md
+++ b/.changeset/plenty-bobcats-know.md
@@ -1,0 +1,5 @@
+---
+'@ithaka/pharos': minor
+---
+
+Fix link not displaying blue outline on focus in Chrome/Safari
--- a/packages/pharos/src/components/link/pharos-link.scss
+++ b/packages/pharos/src/components/link/pharos-link.scss
@@ -21,7 +21,7 @@
     @include mixins.underline;
   }
 
-  display: inline-block;
+  display: block;
 }
 
 :host(:not([href])) #link-element {
--- a/packages/pharos/src/components/link/pharos-link.scss
+++ b/packages/pharos/src/components/link/pharos-link.scss
@@ -20,6 +20,8 @@
   @at-root &.link--alert:active {
     @include mixins.underline;
   }
+
+  display: inline-block;
 }
 
 :host(:not([href])) #link-element {


chrisjbrown (approved):
nice. works for me 👍

daneah (commented):
@NazimHAli can you think of ways that links going from `inline` to `inline-block` could break existing usages?

NazimHAli:
> @NazimHAli can you think of ways that links going from `inline` to `inline-block` could break existing usages?

I was thinking about this and tried to find examples to determine if this was a breaking change. One change that I can think of is with `inline`, the default `width` of the link element is set to `auto`, but with `inline-block`, it will be set to the width of the elements content.

The current `inline`:

Screen Shot 2023-02-02 at 15 24 13

With `inline-block`:

Screen Shot 2023-02-02 at 15 26 14

daneah:
@NazimHAli That's a good one to think about. What I'm interpreting there is that headings that are linked will have their default `block` value forced to `inline-block`, which I don't think is right for headings. Is that accurate?

NazimHAli:
> @NazimHAli That's a good one to think about. What I'm interpreting there is that headings that are linked will have their default `block` value forced to `inline-block`, which I don't think is right for headings. Is that accurate?

It's for the link element, not the header. Another idea is setting it only for links with nested header. It could be the case where the layout isn't rendered instead of all links.

daneah:
@NazimHAli I understand, I was wondering if the heading gets wrapped in the link which affects its display status in turn.

NazimHAli:
@daneah I ran into another oddity, replacing outer style `solid` with `auto` on 243 https://github.com/ithaka/pharos/blob/e50a0f54fe8104b0eae007284767df4d00f46227/packages/pharos/src/utils/scss/_mixins.scss#L241-L246

Renders a different outline, makes me wonder if there's some other styling issue:

pharos jstor org_storybooks_wc_iframe html_id=components-link--visited-link-heading viewMode=story

chrisjbrown:
What was the problem with `inline-block` for the link element? that looks correct for me with `pharos-heading` `div` `span` `just text` slotted in

daneah:
@chrisjbrown a native link around a native heading behaves differently, keeping the full `block` nature of the heading:

Screen Shot 2023-02-03 at 10 42 39

I am not inclined to deviate from this as it will not match people's general expectations.

chrisjbrown:
> @chrisjbrown a native link around a native heading behaves differently, keeping the full `block` nature of the heading:
> I am not inclined to deviate from this as it will not match people's general expectations.

🤔  i see

chrisjbrown:
playing around here.
https://stackblitz.com/edit/pharos-repro-sandbox-4zb9ny?file=src/App.vue

still not sure why pharos doesn't do this in chrome as is 😕

NazimHAli:
@daneah @chrisjbrown any reason we shouldn't set the default `display` for`#link-element` to `block` instead of `inline-block` here?

daneah:
@NazimHAli I think setting links to `display: block` is even more likely to significantly break (almost all?) current usage, as links are typically `display: inline`.

NazimHAli:
> Unless you're saying the host element will still be `display: inline`?

Yes, that's what I mean. Keep `:host` as is, and update `#link-element` to have `display: block`.

NazimHAli:
@daneah setting just the element as a `block` did not work as I'd hoped, it still affects links in general. The only scalable solutions I can think of right now is to create a variant that sets the display for this use case.

PR #493, opened 2023-02-09
Docs: Update flex attribute docs for links 

NazimHAli:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

Adds extra information for users to use the `flex` attribute when using a nested heading so that it's visually accessible in all browsers.

### DIFF
--- a/packages/pharos/src/components/link/pharos-link.ts
+++ b/packages/pharos/src/components/link/pharos-link.ts
@@ -57,7 +57,8 @@ export class PharosLink extends FocusMixin(AnchorElement) {
   public bold = false;
 
   /**
-   * Indicates if the link should display as a flex container.
+   * Indicates if the link should display as a flex container. Enable this with a link heading so that
+   * the blue outline is rendered on focus.
    * @attr flex
    */
   @property({ type: Boolean, reflect: true })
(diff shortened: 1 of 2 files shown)


packages/pharos/src/components/link/pharos-link.ts
```diff
@@ -57,7 +57,8 @@ export class PharosLink extends FocusMixin(AnchorElement) {
   public bold = false;
 
   /**
-   * Indicates if the link should display as a flex container.
+   * Indicates if the link should display as a flex container. Enable this for a link heading so that
+   * the blue outline is rendered on focus for Chrome and Safari (works as expected on Firefox).
```
daneah, line 61: Is there broader guidance that can be given here? I suspect headings might not be the only thing affected, or if they are someone might need to understand why here.

daneah (changes requested):

daneah (approved):
This looks good, I wonder if there are other broader reasons to use `flex` that aren't captured but we can generalize the guidance as needed later on.

NazimHAli:
> This looks good, I wonder if there are other broader reasons to use `flex` that aren't captured but we can generalize the guidance as needed later on.

I haven't had a chance to review general cases, if it comes to mind will open another PR.

chrisjbrown (approved):

Issue #423, opened 2022-10-11
Toast notification and Focus Management


sirrah-tam:
**Expected behavior**

As a keyboard-only user, I'm expecting to dismiss the Toast notification and have keyboard focus be brought back to the triggering control (last known control that triggered the notification) so I can resume navigating the search results. 

**Actual behavior**

After dismissing the Toast notification, focus is lost and is reset to the beginning of the page. I now have to navigate all of way back to where I previously was in order to resume navigation.

**Steps to reproduce the issue (an example)**

1. From search results page, save an item to your workspace.
2. Observe keyboard focus on Toast notification and navigate forward (Tab) to the dismiss button
3. Activate it by pressing Enter
4. Observe keyboard focus after pressing Tab to go to next focusable element.
5. Return to previous location

**Screenshots or code**

<img width="1166" alt="Toast notification workspace" src="https://user-images.githubusercontent.com/6874453/195193372-585886fe-238a-49a5-bc1c-51985701cd81.png">
<img width="1062" alt="Focus after dismissing toast" src="https://user-images.githubusercontent.com/6874453/195193359-0c471795-ecf7-4e16-b87e-88c72f3b49e7.png">

**Pharos version**

Production

**Your environment**

- OS: macOS, Windows
- Browser: Chrome, Firefox, Safari
- Version: 97.0, 105.0.2, 16.0

**Additional information**

Developers can use the element.focus() method to programmatically set focus. Where focus should be placed may be different in certain situations, but this is similar to how modal dialogs work and when dismissed focus should be returned to the control that initially triggered the modal.

This also impacts screen reader users as focus is moved to the Toast notification and similarly when dismissed focus is lost and either is placed randomly on the page or back to the beginning of the page.

There may also be additional accessibility considerations:
- Possible aria-live region to announce content and not rely on pulling users focus
- Persistent vs temporary notifications

brentswisher:
Just a note I thought might help with some additional info. I vaguely remembered the WordPress editor had some [similar accessibility issues](https://github.com/WordPress/gutenberg/issues/16549) when adding a "toast" component. I took a look and while it is still unsolved in that project, there was some interesting reading linked I thought I might surface here in case it was helpful at all:
 - Bootstrap's toast [accessibility notes](https://getbootstrap.com/docs/4.2/components/toasts/#accessibility)
 - PR in Google's "Web Stories" product on [making toasts accessible](https://github.com/GoogleForCreators/web-stories-wp/pull/9630)
[...]

PR #503, opened 2023-03-22
Toast: allow toast to return focus to consumer defined elements after close

jialin-he:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Relates to #423 

Allow toaster to accept consumer defined elements, and try to focus on it after toast is closed

Create property `returnElements` and set focus to the elements in it.

### DIFF
--- a/packages/pharos/src/components/toast/pharos-toaster.ts
+++ b/packages/pharos/src/components/toast/pharos-toaster.ts
@@ -1,5 +1,4 @@
 import { PharosElement } from '../base/pharos-element';
-import { property } from 'lit/decorators.js';
 import { html } from 'lit';
 import type { TemplateResult, CSSResultArray } from 'lit';
 import { toasterStyles } from './pharos-toaster.css';
@@ -28,12 +27,7 @@ import { v4 as uuidv4 } from 'uuid';
  * @listens pharos-toast-open
  */
 export class PharosToaster extends PharosElement {
-  /**
-   * The elements to return focus to after closing the toast
-   * @attr returnElements
-   */
-  @property({ type: Array, reflect: false })
-  public returnElements: Array<HTMLElement> = [];
+  private returnElements: Array<HTMLElement> = [];
 
   constructor() {
     super();


packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -85,6 +94,12 @@ export class PharosToaster extends PharosElement {
     if (toast) {
       this.removeChild(toast);
     }
+    for (const element of this.returnElements) {
+      if (element) {
+        element.focus();
+        break;
+      }
```
daneah, line 101: It might feel better to move this to its own named function, and then you can use `return` instead of `break`.

I think we also need a fallback, returning focus to the main content of the page if no elements were specified or found. Unless that's the default behavior?

jialin-he: Originally I was thinking the consumer will pass in the main content of the page in the array as a backup plan, and that's why it takes the input as an array.

chrisjbrown: `if (element)` doesn't seem to be validating a whole lot here other than that `element` is truthy.

chrisjbrown: this might be a more useful check to validate the node is still present in the document
https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected

packages/pharos/src/components/toast/pharos-toaster.test.ts
```diff
@@ -10,6 +10,7 @@ describe('pharos-toaster', () => {
     const event = new CustomEvent('pharos-toast-open', {
       detail: {
         content: 'I am a toast',
+        returnElements: [],
```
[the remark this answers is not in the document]
jialin-he: Updated the tests!

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -79,12 +88,22 @@ export class PharosToaster extends PharosElement {
     }
   }
 
+  private _focusOnReturnElements(returnElements: Array<HTMLElement>): void {
+    for (const element of returnElements) {
+      if (element && element.isConnected) {
```
chrisjbrown, line 93: can't remember if pharos uses it too much but you could use optional chaining here

```suggestion
      if (element?.isConnected) {
```

packages/pharos/src/components/toast/pharos-toaster.ts
daneah: @jialin-he because consumers don't currently pass this prop at all, I suspect there will be a period where no elements are being passed and we will want to maintain the current behavior in the interim. Does this code already do that?

jialin-he: Yes, if no return element is passed, it maintains the current behavior.

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -27,6 +28,13 @@ import { v4 as uuidv4 } from 'uuid';
  * @listens pharos-toast-open
  */
 export class PharosToaster extends PharosElement {
+  /**
+   * The elements to return focus to after closing the toast
+   * @attr returnElements
+   */
+  @property({ type: Array, reflect: true })
```
daneah, line 35: I don't think we want to reflect this as an attribute in the DOM—it will end up rendering something like `returnelements="[Object, Object, ...]"` which you can sort of see happening in the DOM diff at the bottom of [this page](https://www.chromatic.com/test?appId=60919c26122bd50039b34644&id=641c72d990fd1419d4831bd3).

jialin-he: updated!

packages/pharos/src/components/toast/pharos-toast.wc.stories.jsx
daneah, line 1: There seems to be a whitespace inconsistency on several lines in this file.

packages/pharos/src/components/toast/pharos-toaster.ts
```diff
@@ -27,6 +28,13 @@ import { v4 as uuidv4 } from 'uuid';
  * @listens pharos-toast-open
  */
 export class PharosToaster extends PharosElement {
+  /**
+   * The elements to return focus to after closing the toast
+   * @attr returnElements
+   */
+  @property({ type: Array, reflect: false })
+  public returnElements: Array<HTMLElement> = [];
+
```
daneah, line 37: Now I'm wondering—is this even a property? It seems like the return elements are communicated only through the custom event detail object. Is that right? If so, this can be removed.

jialin-he: ahhhhhh you are right! Updated!

daneah (approved):

PR #508, opened 2023-04-01
 Update the missing href value

### LINKED ISSUE #507: Site: Example Link renders as a button due to missing `href`
**Expected behavior**

The basic example for the Link should probably render an anchor tag, per the principle of least surprise.

**Actual behavior**

The example renders as a button tag because it doesn't have an `href` value.

1. Visit the [component page](https://pharos.jstor.org/components/link)
2. Inspect the **I am a link** link example
3. Observe the `<button>` in its shadow DOM

**Screenshots or code**

Screen Shot 2023-03-31 at 11 04 31 AM

**Additional information**

An `href` value of `#` is probably appropriate.

Yasir761:

fixes #507 

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

**Additional context**
Add any other context here.

### DIFF
--- a/packages/pharos/src/components/link/pharos-link.ts
+++ b/packages/pharos/src/components/link/pharos-link.ts
@@ -115,6 +115,7 @@ export class PharosLink extends FocusMixin(AnchorElement) {
   protected override render(): TemplateResult {
     return this.href !== undefined
       ? html`<a
+          href = "#"
           id="link-element"
           class="${classMap({
             [`link--alert`]: this._alert,
@@ -133,6 +134,7 @@ export class PharosLink extends FocusMixin(AnchorElement) {
         >`
       : html`<button
           id="link-element"
+          href = "#"
           class="${classMap({
             [`link--alert`]: this._alert,
             [`link--hover`]: this._hover,


daneah:
@Yasir761 thank you for contributing! The issue you're referencing is actually a problem only with the site and the example components rendered there, rather than an issue with the component itself. The component should not change, here—just the props being supplied to it in the example on the site!

daneah (changes requested):
@Yasir761 thank you for contributing! The issue you're referencing is actually a problem only with the site and the example components rendered there, rather than an issue with the component itself. The component should not change, here—just the props being supplied to it in the example on the site!

daneah:
@Yasir761 are you planning to address the requested changes?

Yasir761:
> thank you for contributing! The issue you're referencing is actually a problem only with the site and the example components rendered there, rather than an issue with the component itself. The component should not change, here—just the props being supplied to it in the example on the site!

Yes, I'm not able to find the prop where I can pass it.

Also, sorry for being late reply.

PR #535, opened 2023-05-26
fix(tests): update wait for sidenav animation to be less flaky

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Upon reviewing several of the flaky runs of unit tests, the `resets its slide status when going back to a viewport above 1055px` was the frequent culprit. I suspect this is because the animation doesn't finish after the next frame much of the time, so using `aTimeout` to wait instead might be more robust. We use `aTimeout` regularly throughout the tests. We can increase the timeout value if we continue to see flakiness out of it.

Replace `await nextFrame()` with `await aTimeout(1)` in the flaky test.

### DIFF
--- a/packages/pharos/src/components/sidenav/pharos-sidenav.test.ts
+++ b/packages/pharos/src/components/sidenav/pharos-sidenav.test.ts
@@ -1,4 +1,4 @@
-import { fixture, expect, nextFrame } from '@open-wc/testing';
+import { aTimeout, fixture, expect } from '@open-wc/testing';
 import { html } from 'lit/static-html.js';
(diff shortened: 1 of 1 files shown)


daneah:
I saw failures at all of 1, 10, and 100ms, but at 1000ms I've run the tests thrice and seen ✅ all the way through.

mtorres3 (approved):

PR #556, opened 2023-07-08
chore: update dependencies

daneah:
**This change:** (check at least one)

- [ ] Adds a new feature
- [ ] Fixes a bug
- [x] Improves maintainability
- [ ] Improves documentation
- [ ] Is a release activity

**Is this a breaking change?** (check one)

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Title of this pull request clear, concise, and indicative of the issue number it addresses, if any?
- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?
- [ ] Component status page up to date?

PR #558, opened 2023-07-09
Use :focus-visible for most interactive elements

### LINKED ISSUE #175: Accessibility: Investigate use of `:focus-visible` for focus states
**The problem**

Focus states provide valuable feedback to keyboard users, low vision users, and more. However, they can be heavy-handed for cases where a user is already actively clicking on an element, as an example.

**The solution**

Use `:focus-visible` in judicious ways to make Pharos elements show the focus outline only when necessary.

**Alternatives considered**

Leaving the focus states as they are could be alright, but it provides unnecessary motion and can draw attention from the task at hand.

**Additional information**

See [this article](https://matthiasott.com/notes/focus-visible-is-here) for a good overview.

daneah:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No (strictly speaking no API change, but perhaps warrants putting into v13?)

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Resolves #175 

This change uses `:focus-visible` instead of `:focus` for most interactive elements. Notably, it leaves form controls alone, because the focus styling for those feels quite natural as compared to others.

Where buttons, tabs, and links used to have intrusive boxes placed around them on focus for all users, they will now only display those focus outlines when navigating with keyboards or other input devices. Direct mouse pointer input does not create the outline.

The expectation is that this **increases** accessibility for the set of users using mouse pointers, creating less visual noise and motion, while **maintaining** accessibility for all other users. MDN does call out that [there may be cognition
considerations](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#cognition) for the disparity between navigation modes, which I'm not sure how we might best understand more.

This change also provides a fallback for those browsers that don't yet support `:focus-visible` (Safari < 15 at all, and Safari < 15.4 by default) which is to maintain the current behavior of focus outlines all the time.

### DIFF
--- a/.storybook/theme.js
+++ b/.storybook/theme.js
@@ -6,7 +6,6 @@ import {
   PharosColorMarbleGray20,
   PharosFontFamilyBody,
   PharosColorInteractivePrimary,
-  PharosColorInteractiveSecondary,
   PharosColorTextBase,
   PharosColorTextWhite,
 } from '../packages/pharos/lib/styles/variables.js';
@@ -16,7 +15,7 @@ export default create({
   base: 'light',
 
   colorPrimary: PharosColorInteractivePrimary,
-  colorSecondary: PharosColorInteractiveSecondary,
+  colorSecondary: PharosColorInteractivePrimary,
 
   // UI
   appBg: PharosColorWhite,
--- a/packages/pharos/src/utils/scss/_mixins.scss
+++ b/packages/pharos/src/utils/scss/_mixins.scss
@@ -239,9 +239,16 @@
 /// @access public
 /// @group interaction
 @mixin interactive-focus {
-  &:focus {
+  &:focus-visible {
     outline: 2px solid var(--pharos-color-focus);
     outline-offset: 2px;
+  }
+
+  @supports not selector(:focus-visible) {
+    &:focus {
+      outline: 2px solid var(--pharos-color-focus);
+      outline-offset: 2px;
+    }
   }
 }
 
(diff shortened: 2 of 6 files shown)


.storybook/theme.js
```diff
@@ -16,7 +15,7 @@ export default create({
   base: 'light',
 
   colorPrimary: PharosColorInteractivePrimary,
-  colorSecondary: PharosColorInteractiveSecondary,
+  colorSecondary: PharosColorInteractivePrimary,
```
daneah, line 18: Fixes a color bug with the Storybook theme I changed in #556...

packages/pharos/src/utils/scss/_mixins.scss
```diff
@@ -239,10 +239,17 @@
...
     outline-offset: 2px;
   }
+
+  @supports not selector(:focus-visible) {
+    &:focus {
+      outline: 2px solid var(--pharos-color-focus);
+      outline-offset: 2px;
+    }
+  }
 }
```
daneah, line 253: This one affects buttons, links, and their friends in sidenavs, dropdown menus, and so on.

daneah:
@fswlee @sirrah-tam please have a look at [the Storybook](https://60919c26122bd50039b34644-bxveyuttub.chromatic.com/) for this change and see if this feels ballpark.

sirrah-tam:
This looks good to me! I've been trying to find any places where it may cause issues but am coming up short 😃 Also, took a look at the issues around cognition and consistency, but I feel like in our case the change isn't drastic enough to cause concern. Thank you for putting this forward, I think its a great enhancement!

PR #564, opened 2023-07-19
Image card: update image card error variant text

mariadevadoss:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Update Image card error variant text

Change the image card image unavailable text from preview not available to Image preview not available

**Additional context**
Add any other context here.

### DIFF
--- a/packages/pharos/src/components/image-card/pharos-image-card.test.ts
+++ b/packages/pharos/src/components/image-card/pharos-image-card.test.ts
@@ -814,9 +814,6 @@ describe('pharos-image-card', () => {
   it('renders the image preview not available in error state ', async () => {
     component.error = true;
     await component.updateComplete;
-
-    const imagePreview = component.renderRoot.querySelector('.unavailable-text');
-    expect(imagePreview).to.exist;
-    expect(imagePreview?.innerHTML).equals('Image preview not available');
+    expect(component.renderRoot.textContent).to.contain('Image preview not available');
   });
 });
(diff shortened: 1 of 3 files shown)


mtorres3 (approved):

daneah (approved):

jialin-he (approved):

daneah (changes requested):
It seems like we don't have a unit test that hits this case, or at least not one that asserts this text is rendered. Is that something you can look into @mariadevadoss?

packages/pharos/src/components/image-card/pharos-image-card.test.ts
```diff
@@ -810,4 +810,13 @@ describe('pharos-image-card', () => {
...
   });
+
+  it('renders the image preview not available in error state ', async () => {
+    component.error = true;
+    await component.updateComplete;
+
+    const imagePreview = component.renderRoot.querySelector('.unavailable-text');
+    expect(imagePreview).to.exist;
+    expect(imagePreview?.innerHTML).equals('Image preview not available');
+  });
```
daneah, line 821: We typically take a user-first testing approach, so avoid selecting things by their DOM contents when possible. I wonder if we can just check if `component.renderRoot` contains `'Image preview not available'` in some way without knowing about `.unavailable-text`?

mariadevadoss:
```
    component.error = true;
    await component.updateComplete;
    expect(component.renderRoot.textContent).to.contain('Image preview not available');
```
would the modifications above be suitable?

daneah: Yeah I think that is an improvement!

daneah (approved):

PR #566, opened 2023-07-25
Icons: add thumbs up down icons

rehanabbasi:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Adds thumbs up/down icons


daneah:
@rehanabbasi you'll need to remove the enclosing `<svg>` as those get added dynamically in the code; see the linting job failure for more detail. You should be able to run `yarn lint` locally to see this as well.

chrisjbrown (approved):

chrisjbrown:
@rehanabbasi do you need to supply filled and unfilled icons? or can the consumer decide that?

rehanabbasi:
> @rehanabbasi do you need to supply filled and unfilled icons? or can the consumer decide that?

@chrisjbrown I had specified unique icon names for each of those variants!

chrisjbrown:
> > @rehanabbasi do you need to supply filled and unfilled icons? or can the consumer decide that?
> 
> @chrisjbrown I had specified unique icon names for each of those variants!

I mean to say
are those variants necessary?
would it be better to allow the consumer to style those icons

rehanabbasi:
> > > @rehanabbasi do you need to supply filled and unfilled icons? or can the consumer decide that?
> > 
> > 
> > @chrisjbrown I had specified unique icon names for each of those variants!
> 
> I mean to say are those variants necessary? would it be better to allow the consumer to style those icons

Understood - so I think we already have filled and unfilled icon example in the font set (`checkmark-filled-circle
` and `checkmark-inverse` ) where we provide separate icons for each filled and unfilled variant - so was following that!

chrisjbrown (approved):

daneah:
@rehanabbasi @chrisjbrown yeah the key difference here being that these icons actually change nature when filled or not, instead of just changing color. I think it makes sense as well.

daneah (approved):

chrisjbrown:
> @rehanabbasi @chrisjbrown yeah the key difference here being that these icons actually change nature when filled or not, instead of just changing color. I think it makes sense as well.

yeah when i actually looked at the icon didn't seem like any way to get that desired fidelity via the consumer. looks good now

PR #572, opened 2023-07-28
Toast: Prevent error being thrown if opened without returnElements setting

### LINKED ISSUE #571: Toaster component throws errors when Toast is closed if a returnElements is not provided in the calling event
**Expected behavior**
There would be no error if returnElements is provided because it defaults to an empty array

**Actual behavior**
The following error is emitted:

```javascript
pharos-toaster.js:71 Uncaught TypeError: returnElements is not iterable
    at PharosComponent._focusOnReturnElements (pharos-toaster.js:71:1)
    at PharosComponent._closeToast (pharos-toaster.js:84:1)
    at pharos-toast.js:101:1
    at debounce.js:10:1
```

1. Create a toast/toaster by emitting a custom event as detailed in the documentation without including a return element:
```javascript
const event = new CustomEvent("pharos-toast-open", {
        detail: {
            content: "Some problem has occurred",
            status: "error",
        },
    });
    document.dispatchEvent(event);
```
2. Trigger toast to show and then close
3. See error in browser console

**Screenshots or code**
I suspect [this section](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/toast/pharos-toaster.ts#L64) is always overwriting the default, even when it isn't undefined in the event.

```javascript
    const { content, status, id, indefinite, returnElements } = (<CustomEvent>event).detail;
    this.returnElements = returnElements;
```

**Pharos version**
12.19.2

**Your environment**
- OS: macOS Ventura
- Browser: Chrome 
- Version: 114.0.5735.248

brentswisher:

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [X] Changeset added?

Fixes #571 by preventing the default toaster `returnElements` value from being overwritten to `undefined` if a toast event is created without specifying it in the `pharos-toast-open` event.

If the `returnElements` coming from the event details is `undefined` it will be defaulted to an empty array


chrisjbrown (approved):

daneah (commented):
Is there a test that could exercise the bug?

brentswisher:
>Is there a test that could exercise the bug?

I updated the existing mock to not pass an explicit `returnElements` in, before it was passing `[]` So, that is now testing the bug without adding an explicit test. 

I changed the test first, saw it failing, made my change, saw it passing so I do think it is covering the changes.  

Not opposed to adding one though, if you think that would be better?

daneah:
> Not opposed to adding one though, if you think that would be better?

That seems sufficient!

PR #308, opened 2022-02-09
Button: fix styling of buttons with on-background within Vue3 apps

michael-iden:
**This change:** (check at least one)

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves maintainability
- [ ] Improves documentation
- [ ] Is a release activity

**Is this a breaking change?** (check one)

- [X] Yes
- [ ] No

**Is the:** (complete all)

- [X] Title of this pull request clear, concise, and indicative of the issue number it addresses, if any?
- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

**What does this change address?**
Partially addresses #307 

**How does this change work?**
Changes the `on-background` to `is-on-background`

**Additional context**
Yalc'd this change over the the Vue3 app and confirmed the button displays correctly when prefixed with `is-`

PR #578, opened 2023-08-04
Global: update onBackground to isOnBackground

### LINKED ISSUE #307: Update on-background attribute for ease of use in Vue 3
**Expected behavior**
When `on-background` attribute is applied to the button, the color scheme of the button is properly reflected

**Actual behavior**
`on-background` attribute is removed from the DOM, likely by Vue 3, possibly because it starts with `on*` and Vue 3 interprets that to be an event listener.

1. Create pharos button with `on-background` set on the element
2. Load up the app and see that the `on-background` is removed from the DOM
3. Button does not appear as we would expect

**Screenshots or code**
I tried making a codepen/codesandbox/stackblitz but each environment was having various issues. I'll keep trying to get a working example.

https://user-images.githubusercontent.com/13315416/153255838-d599f486-0e0d-4917-afa8-1e81b7e44038.mov

**Pharos version**
11.2.1

**Your environment**

- OS: macOS
- Browser: Chrome, Safari, Firefox
- Version: 98, 15.3, 92.0

**Additional information**
I couldn't find anything in the Vue3 docs where this was happening but the button works perfectly well in the same code on a Vue2 app.

mtorres3:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Fixes #307 
Closes #308


packages/pharos-site/src/components/statics/color/ColorCombos.tsx
```diff
@@ -1,6 +1,8 @@
 import type { FC, CSSProperties } from 'react';
 import { colorBox } from './ColorCombos.module.css';
-import tokens from '@ithaka/pharos/lib/styles/tokens';
+// import tokens from '@ithaka/pharos/lib/styles/tokens.js';
+import tokens from '../../../../../pharos/lib/styles/tokens';
```
eslawski, line 4: So it looks like we are importing the tokens from  the `pharos` source code directly now instead of using the already installed `@ithaka/pharos` package. Is that desired? It creates tighter coupling between `@ithaka/pharos` and `@ithaka/pharos-site`, but it allows us to bridge breaking changes like this in one PR.

If this is desired It's almost like we should uninstall `@ithaka/pharos` from `@ithaka/pharos-site` completely.

If it's not desired, then I wonder if we should leave `@ithaka/pharos-site` alone for now, and circle back to upgrade `@ithaka/pharos` to v13 once it's available. To me this feels like the desired approach, but wanted to raise the point.

mtorres3: @eslawski Without this change to the import, the feature branch kept failing to build. However seeing as we were previously using `@ithaka/pharos`, I was a little hesitant to move forward with this change without a second set of eyes. I think this should work but wanted your opinion as well 👍🏻

eslawski: So my inclination is that we should not be modifying `@ithaka/pharos-site` at this time. This is because `@ithaka/pharos-site` installs `@ithaka/pharos`. So if we focus on only making the updates to `@ithaka/pharos`, we can come back to `@ithaka/pharos-site` later once v13 is released.

eslawski: I might request a sanity check this approach though from other @ithaka/pharos-maintainers

mtorres3: Aha! Makes sense. So revert the `isOnBackground` changes in `/pharos-site/` back to `onBackground` and let it get updated from `@ithaka/pharos/` instead? @eslawski

eslawski: Ya, more specifically I am suggesting we revert any and all changes made in `/pharos-site/`. That I understand will be something we come back and do once the next major version of pharos is cut.

packages/pharos-site/static/guidelines/button.docs.mdx
```diff
@@ -306,8 +306,8 @@ import ellipsesExample from '@images/components/button/buttons_ellipses-example.
   <PageSection title="Buttons on dark background" subSectionLevel={1}>
     <p>
       When a button is used on a darker background (e.g., black or marble-gray-20), use the
-      on-background attribute. The four variants described above can all be used on dark backgrounds
-      and follow the same guidelines.
+      is-on-background attribute. The four variants described above can all be used on dark
```
chrisjbrown, line 309: do we want to hold of on changes to the `pharos-site` until the release is complete and the site is updated to 14?

shoupeva-ithaka (approved):

packages/pharos-site/static/guidelines/button.docs.mdx
mtorres3: Nice catch! Definitely want to remove that 👍🏻

chrisjbrown (approved):

PR #582, opened 2023-08-12
Storybook: remove TypeScript watch from commands

daneah:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [ ] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

I added this in #579 hoping it would automatically watch TypeScript files when running Storybook locally, but unfortunately this seems to interfere with certain things like the loading of icons.

You can achieve this yourself by running e.g. `yarn storybook:wc:dev` and `yarn watch:tsc` in separate processes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,14 +27,14 @@
     "storybook:wc": "yarn storybook:wc:start",
     "storybook:wc:start": "storybook dev -c .storybook/wc -p 9000",
     "prestorybook:wc:dev": "yarn storybook:setup",
-    "storybook:wc:dev": "concurrently \"yarn:storybook:wc:start\" \"yarn:watch:tsc\" \"yarn:watch:styles\" --raw --kill-others",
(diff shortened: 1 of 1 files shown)


SMQuazi (approved):
I haven't tried it it myself but would `&&` work? or perhaps [wait-on](https://www.npmjs.com/package/wait-on) ?

daneah:
@SMQuazi `&&` doesn't work because the commands involved don't complete, and the reason `concurrently` doesn't work here is maybe not something I understand well but my sense is that it's an order-of-operations thing. It's possible `wait-on` might do something useful, but because it's currently broken and requires just the two commands to achieve the desired effect I want to at least ship this and maybe we can open an issue to look at that!

mtorres3 (approved):

PR #594, opened 2023-08-28
a11y revamp: Pharos Buttons

sirrah-tam:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [x] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

This updates the Pharos Button to allow consumers to pass in attributes (listed below) following a specific naming convention (a11y-) that assign the appropriate ARIA attribute to the inner `<button>` element. This addresses the issues with custom web components and the shadow DOM and use of ARIA not being properly being parsed by assistive technologies.

- `a11y-label` = `aria-label` (String)
- `a11y-expanded` = `aria-expanded` (true, false, undefined (default))
- `a11y-pressed` = `aria-pressed` (true, false, mixed, undefined (default))
- `a11y-haspopup` = `aria-haspopup` (true, menu, listbox, tree, grid, dialog, false (default))
- `a11y-disabled` = `aria-disabled` (true, false, undefined (default))

Linked issue #642 

Consumers can add these attributes along with the proper value to the Pharos Button and the appropriate ARIA attribute will be added to the underlying `<button>` element.

**Additional context**
This is a breaking change. The previous versions of Pharos has the `label` attribute that would apply the `aria-label` attribute to the button. This will need to be updated. Additionally, all uses of ARIA listed above will need to be updated accordingly. 

This is an effort to address [ADR #0009](https://github.com/ithaka/pharos/blob/develop/docs/architecture/decisions/0009-create-explicit-minimally-exhaustive-accessibility-properties-for-components.md#9-create-explicit-minimally-exhaustive-accessibility-properties-for-components) (#559)


shoupeva-ithaka (approved):

packages/pharos/tokens/assets/icons.json
```diff
@@ -93,6 +93,7 @@
       "search": { "value": "assets/icons/search.svg" },
       "seek-backward": { "value": "assets/icons/seek-backward.svg" },
       "seek-forward": { "value": "assets/icons/seek-forward.svg" },
+      "send": { "value": "assets/icons/send.svg" },
```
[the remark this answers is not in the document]
sirrah-tam: Oop, good catch, I needed to merge some recent changes on this branch. Managed to grab `develop` but forgot `release/13.0.0`.

mtorres3 (approved):

chrisjbrown:
don't see any reason verify test would be failing now... triggered a rerun just to check

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -63,7 +75,6 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
    * Indicates the variant of button.
    * @attr variant
    */
-  @property({ type: String, reflect: true })
```
sirrah-tam: Oops!

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -269,6 +293,10 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
   }
 
   protected override render(): TemplateResult {
+    // Remove in future release once sufficient time elapsed to update naming convention
+    const a11yLabel = this.a11yLabel ?? this.label;
+    const a11yPressed = this.a11yPressed ?? this.pressed;
```
daneah, line 298: Because this will be released in a major version, I don't believe we need to support backward compatibility like this.

sirrah-tam: My thought was it would encourage faster adoption, it's likely there will be a lot of breaking changes of existing code and the overhead to update everything would require a significant lift.

daneah:
Let's figure out to make sure the diff here is up to date, free of conflicts, and doesn't show others' changes—hard to review until then.

packages/pharos/src/components/sidenav/PharosSidenav.react.stories.jsx
```diff
@@ -60,8 +60,8 @@ export const Base = {
             name="search-button"
             icon="search"
             variant="subtle"
-            label="search"
-            isOnBackground
+            a11y-label="search"
+            onBackground
```
daneah, line 64:
```suggestion
            isOnBackground
```

packages/pharos/src/components/toast/pharos-toast-button.ts
```diff
@@ -18,8 +18,8 @@ export class PharosToastButton extends PharosButton {
     super();
     this.icon = 'close';
     this.variant = 'subtle';
-    this.label = 'Dismiss';
-    this.isOnBackground = true;
+    this.a11yLabel = 'Dismiss';
+    this.onBackground = true;
```
daneah, line 22:
```suggestion
    this.isOnBackground = true;
```

daneah (changes requested):

daneah:
There're a few linting warnings/errors that I think I'm confused about as to why they're arising here in your PR. Several are readily fixed, which I already did but accidentally pushed as a new branch rather than back to your fork; I'll fix that, but ultimately there's a test that might need a closer look that I couldn't fix in the couple of minutes I spent.

sirrah-tam:
> There're a few linting warnings/errors that I think I'm confused about as to why they're arising here in your PR. Several are readily fixed, which I already did but accidentally pushed as a new branch rather than back to your fork; I'll fix that, but ultimately there's a test that might need a closer look that I couldn't fix in the couple of minutes I spent.

It looks like the one error also appears in the `release/14.0.0` branch, the one related to the `any` type. Not sure what's what there. But did manage to clean up the earlier errors referencing unknown ARIA after removing the `aria-description` bits.

daneah:
> the one related to the any type

Yeah, that's the one. If that's the case, let's get this merged in after I push up what I've fixed, since it's not new to this change, and we can attack that error on a different timeline.

daneah (approved):

PR #608, opened 2023-09-08
Coach Mark: Component contribution

henryclong:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Coach Mark

The pharos coach mark will offer teams different ways to onboard users to features and functionality.


packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Indicates which side of the base element the coachmark should appear on
+   * @attr side
+   * @type {Side}
+   */
+  @property({ reflect: true })
+  public side: Side = 'bottom';
+
+  /**
+   * Indicates which side of the base element the coachmark should appear on
```
chrisjbrown, line 52: think this comment needs updated

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Dictates how long to wait between coach mark trigger and start of coach mark fade in animation
+   * @attr delay
+   * @type {Delay}
+   */
+  @property({ reflect: true })
+  public delay: Delay = 'short';
+
+  /**
+   * Text content for the modal header
```
chrisjbrown, line 68: i don't think this is a "modal" header

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,166 @@
...
+  constructor() {
+    super();
+    this.setOffset();
+    this.resizeObserver.observe(document.body);
+  }
+
+  private resizeObserver = new ResizeObserver(
+    debounce((entries) => {
+      entries.map((entry: ResizeObserverEntry) => {
+        if (entry.target.tagName.toLowerCase() !== 'body') return;
```
chrisjbrown, line 100: is this check necessary since only the body tag is being observed?

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
henryclong: fixed

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
henryclong: fixed in [9af8d05](https://github.com/ithaka/pharos/pull/608/commits/9af8d05bfa5f9cc3d76e79874fd310c1fa7c1394)

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
henryclong: It is not required, removed in [9af8d05](https://github.com/ithaka/pharos/pull/608/commits/9af8d05bfa5f9cc3d76e79874fd310c1fa7c1394)

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -0,0 +1,199 @@
...
+  height: fit-content;
+  width: fit-content;
+
+  &.coach-mark-side__right {
+    transform: translateY(-50%);
+
+    .coach-mark__content {
+      --pip-left: 0;
+    }
+  }
```
daneah, line 137: There's a fair amount of nesting going on here; the value of BEM is generally not to need to nest things because everything is similar/same specificity. Can this whole section be flattened out a bit?

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.offsetY = (targetRect.top + targetRect.bottom) / 2 - thisRect.y;
+        break;
+      case 'left':
+        this.offsetY = (targetRect.top + targetRect.bottom) / 2 - thisRect.y;
+        this.offsetX = targetRect.left - thisRect.right - this.margin;
+        break;
+      default: // bottom
+        this.offsetX = (targetRect.left + targetRect.right) / 2 - thisRect.x;
+        this.offsetY = targetRect.bottom - thisRect.top + this.margin;
+    }
```
daneah, line 129: In Tooltip we leverage `floating-ui` to do a lot of the math for us; are we opening ourselves to maintenance burden and human error here?

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,165 @@
...
+          >
+            <pharos-button
+              id="close-button"
+              class="coach-mark__close"
+              type="button"
+              variant="${this.variant === 'light' ? 'subtle' : 'overlay'}"
+              icon="close"
+              label="Close"
+              @click="${() => (this.hide = true)}"
+            ></pharos-button>
```
daneah, line 155: This looks really large to me, when looking at the outcome. Do we want a smaller close button? cc @david-corneail

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  protected override render(): TemplateResult {
+    this.setOffset();
+    return html`
+      <div
+        class="coach-mark ${this.delay && this.delay !== 'none' ? `delay-${this.delay}` : ''}"
+        aria-hidden=${this.hide}
+        role="dialog"
+        aria-labelledby="coach-mark-heading"
+        style="transform:translate(${this.offsetX}px,${this.offsetY}px)"
```
daneah, line 140: Similar here re: `floating-ui`

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
david-corneail: @daneah Yes, I agree — we have close-small icon represented in the figma design and we have a close-small icon in the Pharos UI kit for Figma.. however I'm not seeing the close-small icon loaded into storybook.. @henryclong can you see if you can switch to using the close-small icon.. let's touch base when you get a second

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
henryclong: Good call out. By using `floating-ui` I was able to remove the calculations and reduce the amount of styling required for element positioning

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
henryclong: This inline styling has been removed

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: I was able to reduce the amount of nesting, which was also helped by the addition of `floating-ui` mentioned below

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,151 @@
...
+    }).then(({ x, y }) => {
+      Object.assign(this.style, {
+        left: `${x}px`,
+        top: `${y}px`,
+      });
+    });
+  }
+
+  protected override render(): TemplateResult {
+    this.setOffset();
```
chrisjbrown, line 120: does this need called at render if it's already called in the constructor?

daneah: I think the component may get constructed just once, but rendered multiple times?

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -0,0 +1,183 @@
...
+  display: block;
+  position: absolute;
+}
+
+:host([variant='dark']) {
+  --coach-mark-background-color: var(--pharos-button-color-base-overlay-background-base);
+  --coach-mark-border-color: var(--coach-mark-background-color);
+  --coach-mark-heading-color: var(--pharos-color-white);
+  --coach-mark-text-color: var(--pharos-color-marble-gray-80);
+}
```
daneah, line 58: These have typically been defined as component tokens up at the top level with all the other tokens (see `packages/pharos/tokens/components/`). Recently we've been discussing whether to continue using component tokens like that at all, so I'm a little uncertain what to suggest here; if we were to get rid of component tokens at the top level, I like this way of handling the CSS properties. Interested in thoughts from other @ithaka/pharos-maintainers

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -0,0 +1,183 @@
...
+  &::after {
+    --coach-mark-border-width-x2: calc(var(--coach-mark-border-width) * 2);
+
+    background-color: var(--coach-mark-border-color);
+    height: calc(var(--coach-mark-pip-size) + var(--coach-mark-border-width-x2));
+    width: calc(var(--coach-mark-pip-size) + var(--coach-mark-border-width-x2));
+    z-index: -1;
+  }
+
+  pharos-heading {
```
daneah, line 114: Re: BEM please consider adding a class to the nested heading element and using that as an (unnested) selector for this instead.

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -0,0 +1,183 @@
...
+    color: var(--coach-mark-heading-color);
+    padding-right: 24px;
```
daneah, line 116:
```suggestion
    padding-right: var(--pharos-spacing-one-and-a-half-x);
```

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -0,0 +1,183 @@
...
+    animation-name: fadein-long;
+  }
+}
+
+.coach-mark__content {
+  position: relative;
+  background-color: var(--coach-mark-background-color);
+  border: var(--coach-mark-border-width) solid var(--coach-mark-border-color);
+  padding: var(--pharos-spacing-one-half-x) var(--pharos-spacing-three-quarters-x);
+  min-width: 100px;
```
daneah, line 86: @henryclong @david-corneail might a `ch` unit here be more useful? "The coach mark should fit at least N characters wide"

david-corneail: perhaps 30 characters wide, including "blank" spaces.. before the text falls to another line @henryclong @daneah

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: addressed in [7d94f10](https://github.com/ithaka/pharos/pull/608/commits/7d94f10c009a4e59cd1b92f959c5d00555b6ec75)

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: addressed in [7d94f10](https://github.com/ithaka/pharos/pull/608/commits/7d94f10c009a4e59cd1b92f959c5d00555b6ec75)

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: I've gone ahead and updated this to default to 30ch. I've also updated the width attribute of the element to accept units so it's not exclusively tied to px

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: Should I move these values to `packages/pharos/tokens/components/` in the mean time?

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Set minimum width of coach mark
+   * @attr width
+   * @type {string}
+   */
+  @property({ reflect: true })
+  public width = '30ch';
+
+  private computedSide = 'bottom';
```
daneah, line 87: Our convention is typically to prefix private members with underscore.

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -0,0 +1,183 @@
...
+  --pip-left: 100%;
+}
+
+.coach-mark-side__top .coach-mark__content {
+  --pip-top: 100%;
+}
+
+.coach-mark-side__bottom .coach-mark__content {
+  --pip-top: 0;
+}
```
daneah, line 146: With BEM we might also consider something like `.coach-mark__side--right` since the "side" feels like the element and _which side it is_ feels like the modifier.

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
daneah: If it isn't too much trouble, I think that's the right move for now. Be aware of [this decision](https://github.com/ithaka/pharos/blob/develop/docs/architecture/decisions/0008-evolve-tokens-in-backward-compatible-manner.md) that means I think you'll need to provide the more raw values above as fallbacks when using the tokens in this component.

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: addressed in [2e541ad](https://github.com/ithaka/pharos/pull/608/commits/2e541ad604a4cce7e0990e07409136c19bbaf632)

packages/pharos/src/components/coach-mark/pharos-coach-mark.ts
henryclong: addressed in [497a642](https://github.com/ithaka/pharos/pull/608/commits/497a6429f8859b2f01eb4f83d8449c05977817da)

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
henryclong: I've moved the pharos tokens to `packages/pharos/tokens/components/coach-mark.json`, but I'm using them to populate the light/dark color variables here so I don't need to refactor the color theme switching

packages/pharos/src/components/coach-mark/pharos-coach-mark.scss
```diff
@@ -37,24 +37,20 @@
...
 
 :host([variant='dark']) {
-  --coach-mark-background-color: var(--pharos-button-color-base-overlay-background-base);
-  --coach-mark-border-color: var(--coach-mark-background-color);
-  --coach-mark-heading-color: var(--pharos-color-white);
-  --coach-mark-text-color: var(--pharos-color-marble-gray-80);
+  --coach-mark-background-color: var(--pharos-coach-mark-color-background-dark);
+  --coach-mark-border-color: var(--pharos-coach-mark-color-border-dark);
+  --coach-mark-heading-color: var(--pharos-coach-mark-color-heading-dark);
+  --coach-mark-text-color: var(--pharos-coach-mark-color-text-dark);
```
daneah: Sorry, it seems I wasn't too clear before; I think you will want to remove these custom properties now, and have CSS below like the following:

```css
... {
  background: var(--pharos-coach-mark-background-color, var(/* the raw value from before */));
}
```

daneah (approved):

chrisjbrown (approved):

PR #613, opened 2023-09-13
Sheet: Add overflow for long content and remove footer

jialin-he:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

Fix the sheet to allow it scroll with long content.

Before:
image

After:
image

Add overflow and other css constraints.
Also remove the header and footer from sheet to match what we did for popover.

### DIFF
--- a/packages/pharos/src/components/sheet/pharos-sheet.scss
+++ b/packages/pharos/src/components/sheet/pharos-sheet.scss
@@ -74,19 +74,7 @@
   outline: 0;
   transition: height var(--pharos-transition-base);
   box-shadow: 0 1px 2px 0 rgb(18 18 18 / 0.3), 0 -4px 8px 3px rgb(18 18 18 / 0.15);
-
-  .sheet__wrapper {
-    padding: 0 var(--pharos-spacing-one-and-a-half-x);
-    display: flex;
-    flex-direction: column;
-    height: calc(100% - 67px);
-  }
-
-  @include mixins.until(small) {
-    .sheet__wrapper {
-      padding: 0 20px;
-    }
-  }
+  contain: layout;
 
   .sheet__handle {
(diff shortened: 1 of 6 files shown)


packages/pharos/src/components/sheet/pharos-sheet.scss
```diff
@@ -77,6 +77,9 @@
 
   .sheet__wrapper {
     padding: 0 var(--pharos-spacing-one-and-a-half-x);
+    display: flex;
+    flex-direction: column;
+    height: calc(100% - 67px);
```
daneah, line 82: Where does this `67px` value arise from?

chrisjbrown: is specifying the height necessary? doesn't seem to effect it if i remove it
Screenshot 2023-09-14 at 4 05 08 PM

chrisjbrown: ah nvm. i see it's needed for overflow scrolling 🤔

chrisjbrown: does replacing the `height` with `overflow-y: scroll` achieve the same result?
Screenshot 2023-09-14 at 4 20 37 PM

jialin-he: Thank you for the advice! After discussing with Lori we now decide to actually remove the footer from the sheet (at least for now) to match the treatment we have in the popover. So this problem naturally solves.

chrisjbrown (approved):

PR #628, opened 2023-10-05
A11y revamp: Pharos buttons (non-breaking change)

sirrah-tam:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

This updates the Pharos Button to allow consumers to pass in attributes (listed below) following a specific naming convention (a11y-) that assign the appropriate ARIA attribute to the inner <button> element. This addresses the issues with custom web components and the shadow DOM and use of ARIA not being properly being parsed by assistive technologies.

    a11y-label = aria-label (String)
    a11y-description = aria-description (String)
    a11y-expanded = aria-expanded (true, false, undefined (default))
    a11y-pressed = aria-pressed (true, false, mixed, undefined (default))
    a11y-haspopup = aria-haspopup (true, menu, listbox, tree, grid, dialog, false (default))

How does this change work?
Consumers can add these attributes along with the proper value to the Pharos Button and the appropriate ARIA attribute will be added to the underlying <button> element.

**Additional context**
This is not a breaking change as consumers are still able to utilize the `label` attribute, but does add a browser warning about the deprecated `label` attribute. The major release (v14) will include PR #594 and will remove all backwards compatibility as we include additional breaking changes. Additionally, all uses of ARIA listed above currently on the Pharos button will need to be updated accordingly to work as expected.


mtorres3 (approved):
Looks good!

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -178,6 +225,16 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
         `${this.variant} is not a valid variant. Valid variants are: ${VARIANTS.join(', ')}`
       );
     }
+
+    // Warn consumers that the label attribute is being deprecated
```
daneah, line 229:
```suggestion
```

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -254,23 +316,30 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
...
+            aria-pressed=${ifDefined(a11yPressed)}
+            aria-expanded=${ifDefined(this.a11yExpanded)}
+            aria-haspopup=${ifDefined(this.a11yHaspopup)}
             @keyup=${this._handleKeyup}
           >
             ${this.buttonContent}
           </a>
         `
       : html`
+          <!-- @ts-ignore -->
```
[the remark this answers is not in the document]
brentswisher: From what I recall , this was because of an issue in `aria-haspopup` with lit-analyser.

I started looking into it but couldn't reproduce the problem locally from their source. That's when I saw that despite recent work, a new package hasn't been release in almost 3 years.

I asked [here](https://github.com/runem/lit-analyzer/pull/175#issuecomment-1765306567) about that, hopefully they can clarify. If they aren't releasing new versions anymore we might have to look into migrating to something else?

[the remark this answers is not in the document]
daneah: Is there a way we can ignore the single problematic attribute with an inline comment rather than the whole tag?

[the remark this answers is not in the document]
brentswisher: Not that I found, although it wasn't an exhaustive search. Per-rule comments didn't work and in the `ts-lint-typescript` documentation it only mentions the broad rule. https://github.com/rictic/ts-lit-plugin#-support-for-ts-ignore-comments-inside-html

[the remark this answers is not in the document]
daneah: May be able to check on this with `lit-analyzer@^2.0.1`.

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -112,12 +124,41 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
   public large = false;
 
   /**
+   * @deprecated
```
shoupeva-ithaka, line 127: I may be incorrect, but does Pharos v14 introduce the new accessibility properties on most components? I guess I ask because if we released this with v14 we could remove the "deprecated" attributes entirely instead of providing backwards compatibility. I imagine at some point or another we would want to remove those attributes which would be a breaking change and a major version bump.

shoupeva-ithaka (approved):

packages/pharos/src/components/button/pharos-button.ts
sirrah-tam: Hey @shoupeva-ithaka, for the v14 release, we'll be removing the backwards compatibility entirely so the deprecated callouts will also be removed. This PR is for the minor release to get those folks who are ready to move to the new attributes a way to do so, while flagging for other folks that it'll be deprecated come the v14 major release.

sirrah-tam: Also, to follow up, when v14 is released it won't be adding the attributes to the components per se, consumers will still need to update the attributes on the current implementation of specific components.

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -243,6 +285,10 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
   }
 
   protected override render(): TemplateResult {
+    // Remove in future release once sufficient time elapsed to update naming convention
```
daneah, line 288:
```suggestion
    // TODO: Remove in future release once sufficient time elapsed to update naming convention
```

daneah (approved):

packages/pharos/src/typings/a11y-attributes.d.ts
```diff
@@ -0,0 +1,17 @@
+export {};
+
+declare global {
+  type AriaHiddenValues = 'false' | 'true' | 'undefined' | undefined;
```
brentswisher, line 4: Could this be update to `AriaHiddenState` just for consistency?

brentswisher (approved):

Issue #590, opened 2023-08-21
Add new components to `initComponents` files


michael-iden:
**The problem**
I spent nearly an hour trying to debug my tests thinking there was a problem with the code, not realizing that I needed to add my new component definition to [this file to initialize components within the unit tests](https://github.com/ithaka/pharos/blob/44cc1a3ee1a494ac48af2371231b8de2f124c9da/packages/pharos/src/test/initComponents.ts).

**The solution**
New components created through the CLI are automatically added to `initComponents` files(s)

**Alternatives considered**
Add a direction to make this addition to the [CLI readme](https://github.com/ithaka/pharos/blob/730d381dd86fffe2543e37e2ef9a367611a997f1/packages/pharos-cli/README.md).

**Additional information**
We also need to add new component registration to the [storybook `initComponents`](https://github.com/ithaka/pharos/blob/4b833fab5e9bb65d0c03619f26cf1875ff16b400/.storybook/initComponents.js), but that felt more obviously necessary when I was visually testing compared to the failing unit tests.

michael-iden:
@satya-achanta-venkata This seems reasonable to me. The concept and your source of inspiration for this change look like the right direction.

daneah:
Addressed in #630

PR #630, opened 2023-10-10
feat(cli): add newly created components created using pharos-cli to initComponents files

satya-achanta-venkata:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

In #590 , After creating new components using `pharos-cli`, we had to make manual changes to `initComponents.[j|t]s` files  to components being recognized by tests. We found a room to automate the step of adding newly created components to `initComponents` files so that developer no need to make manual changes.

For example: If we create new component using `pharos wc yahoo`, new component wil automatically be added to `packages/pharos/src/initComponent.ts` and `.storybook/initComponents.js` files. 

**Additional context**
Try this by following below steps:
1. check out this branch locally
2. run `pharos wc tresleches`
3. You should be seeing newly added components are imported and registered in below two files
- `packages/pharos/src/initComponent.ts`
- `.storybook/initComponents.js`


packages/pharos-cli/src/utils/generation-utils.ts
```diff
@@ -68,3 +68,29 @@ export const addComponentToIndexFile = (
...
+  currentDirectory: FilePath,
+  nameOptions: ComponentNameOptions
+): void => {
+  const componentName = `Pharos${nameOptions.titleCaseName}`;
+  const unitTestsFilePath: FilePath = `${currentDirectory}/packages/pharos/src/test/initComponents.ts`;
+  const storybookFilePath: FilePath = `${currentDirectory}/.storybook/initComponents.js`;
+
+  // import new component in list of imports
+  // and then register new component
+  [unitTestsFilePath, storybookFilePath].map((filePath: FilePath) => {
```
daneah, line 82: Probably better to use `.forEach` here, as you aren't using the returned array from `.map`.

packages/pharos-cli/src/utils/generation-utils.ts
```diff
@@ -68,3 +68,29 @@ export const addComponentToIndexFile = (
...
+    fs.readFile(filePath, 'utf8', function (_, data) {
+      const existingFileContent: string[] = data.split('} from');
+      const newImportStatement =
+        existingFileContent[0] + `  ${componentName}` + ',' + '\n' + '} from';
+      const updatedFileContent = newImportStatement + existingFileContent[1];
+      fs.writeFileSync(filePath, updatedFileContent);
+
+      const existingRegisterComponents: string[] = updatedFileContent.split(']);');
```
daneah, line 90: This `split`ting feels brittle; if the contents of the file being manipulated changes substantially later, this operation could produce a bad file. Maybe it's a low-likelihood thing but something to think a bit about.

packages/pharos-cli/src/cmds/pharos-component.ts
```diff
@@ -71,6 +72,7 @@ const createPharosComponent = async (componentName: ComponentName): Promise<void
   if (!wasError) {
     console.log(colors.yellow(`\nAdding Pharos${titleCaseName} component export`));
     addComponentToIndexFile(currentDirectory, nameOptions);
+    addComponentsToInitFile(currentDirectory, nameOptions);
```
daneah, line 75: Do we want to add another similar log, or generalize the existing one a bit?

packages/pharos-cli/src/utils/generation-utils.ts
satya-achanta-venkata: @daneah  I agree with you. I also feel the same. However, I tried not to use this way of splittin,  but I couldn't come up with better way of doing this. Do you think of any better way like using pure typescript or such ? I would love to use that.

packages/pharos-cli/src/cmds/pharos-component.ts
satya-achanta-venkata: Good idea. Will add another new log that logs the message clearly about adding components to init files.

packages/pharos-cli/src/utils/generation-utils.ts
daneah: The most robust way I know of to alter a file statically is to manipulate and validate an altered [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree). The TypeScript wiki hints at [some ways of doing that](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#re-printing-sections-of-a-typescript-file) using the TypeScript API. This is very possible overkill here; I won't necessarily encourage going down that road.

packages/pharos-cli/src/utils/generation-utils.ts
satya-achanta-venkata: switched to `forEach`. So resolving the conversation.

packages/pharos-cli/src/cmds/pharos-component.ts
satya-achanta-venkata: Added new log with color `green`. Let me know if you feel any other color would be apt @daneah

daneah: I think yellow like the existing log would be good. Green indicates success, but the log is saying something is about to happen, not that it already succeeded.

packages/pharos-cli/src/utils/generation-utils.ts
satya-achanta-venkata: @daneah  I am working on modifying source file using AST,  the links you shared are useful.  So, its just taking little longer than I expected. Hoping to get it done in next two days.

packages/pharos-cli/src/cmds/pharos-component.ts
satya-achanta-venkata: updated log to yellow. So, resolving the conversation.

packages/pharos-cli/src/utils/generation-utils.ts
satya-achanta-venkata: I see that you mentioned, it would be overkill, but I wanna give it a shot. I liked that API a lot, it's interesting.

satya-achanta-venkata: @daneah  after spending good amount of time, I learned that using `ts-morph` package is helpful and it lets us do what we need in lesser code. I got stackblitz project with working function. I added details in README about how I used the script. Let me know your thoughts on this. More explicitly, I am looking for your suggestion on whether we can use `ts-morph` package in our  pharos site.

daneah: Nice find! I think I'd welcome giving this a try, especially as a `devDependencies` in `pharos-cli`.

satya-achanta-venkata: Thanks, Dane. Will give it a shot today.

satya-achanta-venkata: @daneah I am facing this error when running `yarn add -D ts-morph` in `pharos-cli`. I am leaning towards installing the missing package, but wondering am I misunderstanding anything ? Can I get some help here ?

`error An unexpected error occurred: "expected workspace package to exist for \"@typescript-eslint/eslint-plugin\"".
`

packages/pharos-cli/src/utils/initComponentsUtil.ts
```diff
@@ -0,0 +1,28 @@
...
+  const project = new Project();
+  const sourceFile = project.createSourceFile(inputFilePath, code, {
+    overwrite: true,
+  });
+
+  const importStatementsDeclarations = sourceFile.getDescendantsOfKind(
+    SyntaxKind.ImportDeclaration
+  );
+
+  importStatementsDeclarations[0].addNamedImport(componentName);
```
daneah, line 19: This in general is looking great. I think the last brittle piece could be this `[0]` here; if the imports change order, this would break. Is it simple enough to loop through the import declarations and match the one that's importing the components?

I don't want you to have to spend too much longer on this, so I'd also be happy starting with this and maybe opening a new issue to solidify things (and probably to convert the other pieces to use `ts-morph`, since this seems nicer).

daneah (approved):

packages/pharos-cli/src/utils/initComponentsUtil.ts
satya-achanta-venkata: @daneah  I can give it a shot. I like what you mentioned, yes, I can improve it while I still have this PR opened.

packages/pharos-cli/src/utils/initComponentsUtil.ts
```diff
@@ -16,13 +16,22 @@ export async function addNewComponentToInitComponents(
     SyntaxKind.ImportDeclaration
   );
 
-  importStatementsDeclarations[0].addNamedImport(componentName);
+  importStatementsDeclarations.forEach((dec) => {
+    const importClause = dec.getImportClause();
+    if (importClause && importClause.getNamedImports().length > 0) {
```
satya-achanta-venkata, line 21: Appreciate the suggestion above @daneah . I improved the code using syntax kinda instead of elements. I too agree that if we alter the import statements order on the page, script will break. With this change, our script should add required component name exactly where we want.

packages/pharos-cli/src/utils/initComponentsUtil.ts
satya-achanta-venkata: Update to nicer way in latest commit, so resolving this conversation.

Issue #467, opened 2022-12-22
Require descriptions for non-decorative icons without other visible descriptive text


daneah:
**The problem**

Some buttons have only an icon and trigger a tooltip on hover / focus. This may work for sighted users, but is not sufficient to convey meaning for those using assistive technologies.

**The solution**

When an icon appears in a non-decorative context, require that the developer provide a description. Note that this may require property drilling for things like Button whose API only allows the developer to reference an icon by name rather than providing any and all Icon properties directly.

**Alternatives considered**

We discussed the option of making all Icons require the description attribute, but there are valid use cases for decorative icons that don't require descriptive text. It isn't clear whether we intend or already have in practice used icons this way.

We could potentially require a description in all cases and have a `hidden` property or similar to hide decorative icons from being announced, thereby forcing the developer to opt explicitly for creating an icon without a description.

**Additional information**

Once an approach is finalized, we should provide guidance in the appropriate component documentation.

daneah:
The work remaining here after #646 will be to remove the deprecated way of doing this using the `description` property.

brentswisher:
Closed in #679

PR #646, opened 2023-11-13
Add a11y properties to icon component

brentswisher:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [X] Changeset added?

Adds both a a11y-title and and a11y-hidden property to the icon component. This is a replacement for the previous description which is now marked as deprecated and will be removed in the next major release (#467).  It also replaces the use of aria-label with setting a title in the svg of the icon which after talking to @sirrah-tam is the preferred method. 

Adds the a11y-title and a11y-hidden property and maps them to a title and aria-hidden attribute on the icon svg.  The previous functionality of aria-hidden automatically being set if no description being provided is preserved here to enable adding the new attributes without a breaking change. 

**Additional context**
See #467 to track the breaking change that follows this in the next major release.


packages/pharos/src/components/icon/pharos-icon.ts
```diff
@@ -59,20 +84,22 @@ export class PharosIcon extends PharosElement {
...
     return html`
       <svg
         xmlns="http://www.w3.org/2000/svg"
         version="1.1"
         viewBox="0 0 ${size} ${size}"
         class="icon"
         role="img"
-        aria-hidden=${this.description === ''}
-        aria-label=${this.description || ''}
+        aria-hidden=${hideIcon || nothing}
```
daneah, line 97: I think there's better shorthand for this unless I'm missing something:

```suggestion
        ?aria-hidden=${hideIcon}
```

packages/pharos/src/components/icon/storyArgs.ts
```diff
@@ -3,6 +3,8 @@ import { iconNames } from '../../utils/iconNames';
 export const defaultArgs = {
   name: 'checkmark',
   description: '',
+  a11yTitle: 'This is the icon title',
```
sirrah-tam, line 6: This does give the example a way of indicating how the a11yTitle is used, but should we include a more relevant title for the purpose of the icon, eg. "checkmark"?

packages/pharos/src/components/icon/pharos-icon.ts
```diff
@@ -30,13 +30,38 @@ export class PharosIcon extends PharosElement {
   /**
    * A description of what the icon represents
    * @attr description
+   * @deprecated Please use a11yLabel instead.
    */
   @property({ type: String, reflect: true })
   public description = '';
 
+  /**
+   * Indicates the svg title to apply to the icon.
```
sirrah-tam, line 39: I think it might be worthwhile to call out that the title is the SVG accessible name.
```suggestion
"Indicates the title to apply as the accessible name of the icon"
```

packages/pharos/src/components/icon/pharos-icon.ts
```diff
@@ -30,13 +30,38 @@ export class PharosIcon extends PharosElement {
   /**
    * A description of what the icon represents
    * @attr description
+   * @deprecated Please use a11yLabel instead.
```
sirrah-tam, line 33: This should be updated to "Please use a11yTitle instead."

```suggestion
   * @deprecated Please use a11yTitle instead.
```

packages/pharos/src/components/icon/pharos-icon.ts
daneah: [Here are the relevant portion of the docs](https://lit.dev/docs/templates/expressions/#boolean-attribute-expressions) that I think mean the attribute will indeed be removed.

brentswisher: Okay, I think I might see what is happening. When using `?aria-hidden=${hideIcon}` I get a bunch of failing unit tests from other components. That matched what I saw before when trying it. I had thought it was because it was not removing the `aria-hidden` attribute when `hideIcon` it was false. The tests are actually failing because when `hideIcon` is set to true, it is rendering the svg as `aria-hidden=""` which the axe plugin we use doesn't seem to recognize as being hidden.

With the `nothing` syntax `aria-hidden` is either not present or set explicitly to `aria-hidden="true"` which axe does seem to recognize. I'll see if there is a bug report for axe about that or if it is expected and we can go from there, it does seem to be rendering correctly in the browser with either syntax.

sirrah-tam: A lil bit more research, I think that `aria-hidden` needs to have an explicit value set per the [spec](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden) because it's a state and not a boolean attribute. So would need to be "true", "false", or undefined. 

Reference: https://stackoverflow.com/questions/55061083/does-aria-hidden-need-a-value

daneah: @brentswisher that doesn't jibe with the docs or my understanding...something else is perhaps wrong? Or maybe the thing you think is evaluating as falsy is actually still truthy in some way?

@sirrah-tam I think we're almost saying the same thing, but I propose that `aria-hidden` should not be present here unless its value is `true`; to follow the spec exactly the way you're saying would imply to me that every single element in the DOM for which `aria-hidden` can be set would need to have it explicitly set to `aria-hidden="false"` which isn't my understanding.

brentswisher: Yeah, I think the root of this is the fact we are using aria-hidden as a boolean attribute when it isn't.

By using `?` we are explicitly telling Lit "This is a boolean attribute" 

From the [HTML spec](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes):
>The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

So Lit thinks based on the value of `hideIcon` the `aria-hidden` attribute needs to either:
- be present
- not exist 

Importantly, it never sets the value to anything either way, it just includes it or not. 

This matches what I am seeing in storybook (Note the `data-hidden-value` which I added to debug the value of `hideIcon`):
Screenshot 2023-11-29 at 1 40 29 PM

This also causes our accessibility tests to fail in other components: when aria-hidden should be `aria-hidden="true"` it is rendered as just `aria-hidden`

By using `aria-hidden=${hideIcon || nothing}` Lit doesn't treat it as a boolean, so it is correctly outputting `aria-hidden="true"`

I think the mental complication for this is stemming from the fact I specified `a11yHidden` as a boolean property on the component, which translates into `aria-hidden` on the svg, which is not a boolean property.

sirrah-tam: Yep, I think we're saying the same thing, and IMO it makes the most sense to only render `aria-hidden` if it's needed to be set to true. My previous message was mostly trying to say that we shouldn't expect that only having `aria-hidden` without an assigned value (`aria-hidden=""`) to be parsed as a truthy attribute. The Accessibility API will ignore it, and it's why the axe plugin was throwing errors. 

💯 ☝️  @brentswisher

daneah: This makes sense. @brentswisher @sirrah-tam do we want to make `a11yHidden` align with the `aria-hidden` spec for convergence/consistency's sake, and keep the rendering line here as-is?

daneah: The overlap of meaning of JavaScript boolean `true|false` values and `undefined` with the string values `"true"|"false"|"undefined"` makes this challenging. I note that in Button's pressed state we handle everything explicitly as a string, so that they're always truthy (`"true"|"false"|"undefined"` are all truthy, as strings). I feel like that is closest to authoring raw HTML as well; if what you type is what ends up in the browser, most attributes then become either boolean (that is, present or not) or string. 

In Button we codified this even more directly for its pressed state; note that its type is `'true' | 'false' | 'undefined' | undefined` (different than what you have). I feel that the spec is really forcing `aria-hidden` to be a thing whose value is really a string, which gets coerced to a boolean or to a "idk, let the browser decide" behavior. I would vote we stick to that because I'd rather people maintain a mental model of the web than add a layer of indirection and muddy understanding when using these. My two cents.

sirrah-tam (approved):

packages/pharos/src/components/icon/pharos-icon.ts
brentswisher: Pharos Icon a11y excalidraw


I have worked with Mat a bit to come up with this and believe the most recent code matches these expectations. I can also reuse the [ariaHiddenState](https://github.com/ithaka/pharos/pull/628/files#diff-ab0c9cb8869e1f27ee4d0a3c3532056dfee9e98d452e8a8c4cf947f425331159R4) from the button PR once that gets merged into develop for consistency.

daneah: This is very helpful, thank you for documenting!

daneah (approved):

satya-achanta-venkata (approved):

shoupeva-ithaka (approved):

Issue #635, opened 2023-10-12
Upgrade to Yarn berry / Yarn PnP


daneah:
**The problem**

We're currently on Yarn 1, and don't want to get too far behind. Recent versions of Yarn have [several benefits and features](https://yarnpkg.com/migration/overview) that are being used internally that would also benefit this project.

**The solution**

Upgrade to Yarn berry / Yarn PnP

**Additional information**

* Follow the migration guide for getting on Yarn berry
* Follow the "To go further" guide to get on Yarn PnP
* Update user scripts with `pre` and `post` to be run explicitly
* Fix missing and misconfigured dependencies based on build/process errors

I tried this briefly myself, and got just about through all the steps, before finally running into some TypeScript issues—this could be because of some additional missing/incorrect TypeScript configuration/dependencies.

michael-iden:
@daneah I would like to contribute to this in the next few weeks if you also have availability. I don't feel I have the experience with the pharos setup to go it alone with confidence. Related, [I have a likely unmergable PR](https://github.com/ithaka/pharos/pull/650/files#diff-3fec3e7681402b1e4799321a89db6a2292e9ccd08609ab5150102e2c9aff9b47) because of issues running pharos where my global yarn is now 4.x

daneah:
This, as written, was handled in #694—we may eventually want to try out Zero installs, but that's a different story!

PR #650, opened 2023-11-20
Card: Improve accessibility by moving `overlay` slot outside of `<a>` tag

michael-iden:

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No - Although I think its pretty close to a breaking change

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

Fixes a problem where the button inside of the `overlay` slot would not be picked up by the browser when tabbing through clickable elements. I believe this is because the button within the `overlay` was slotted inside of an `anchor` tag

Move overlay slot outside of the link tag

**Additional context**
I am very nervous about this change. This component is far too complex and takes in too many variable elements through slots. Even with tests and the sandbox I have little confidence that it's not actually going to break the UX in at least some browsers.

Also I am struggling to get the hooks to run properly with the shimmed yarn version. My global yarn in `4.0.1` in the node envs 18+ managed by `asdf`. I might look to update the application setup to a later yarn if we can get a moment (#635).


package.json
```diff
@@ -147,5 +139,5 @@
       "path": "packages/pharos/lib/index.js"
     }
   ],
-  "packageManager": "yarn@1.22.19"
+  "packageManager": "yarn@4.0.2"
```
michael-iden, line 142: This isn't right

packages/pharos/src/components/image-card/pharos-image-card.wc.stories.jsx
```diff
@@ -390,6 +390,53 @@ export const WithActionButtonSlot = {
     `,
 };
 
+export const WithOverlayButtonSlot = {
```
daneah, line 393: Can you make sure to duplicate this in the React stories as well?

package.json
daneah: Let's see if we can revert all the Yarn-related items for the moment.

brentswisher:
@michael-iden Wondering what the status of this is, I see it has been open a few months. Is there work needed here? Should it be closed as no longer needed?

michael-iden:
@brentswisher this is still an issue but was unfortunately deprioritized and I haven't gotten back around to updating things. I'll try to get an estimate on when I might be able to circle back on this fix and I'll send you a thread directly where we were tangentially talking about this problem.

brentswisher:
@michael-iden No worries at all, appreciate the work in this, just trying to do a little GitHub cleanup where I can! 😄

PR #679, opened 2024-02-02
Require icon accessibility attributes

### LINKED ISSUE #467: Require descriptions for non-decorative icons without other visible descriptive text

brentswisher:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [X] Yes
- [ ] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [X] Changeset added?

Fixes #467. 
This is the breaking change followup to #646, removing the icon description and requiring an icon to be correctly annotated for screen readers by providing a title or being marked as hidden.

It does the following for the icon component:
- Removes the deprecated `description` attribute entirely
- No longer defaults to hiding the icon if there is no accessible title provided
- Throws an error if an icon is added without being provided a title or being explicitly marked as hidden

**Additional context**
- See [this diagram](https://github.com/ithaka/pharos/pull/646#discussion_r1420728515) for a better understanding of what has changed between v13/v14. This PR represents the end state of that diagram.
- I also updated several downstream tests which were expecting the default for description to be present in their checks.
- There was an icon I missed in #660 in the dropdown navigation so I fixed that too. (Some nice proof this will prevent inaccessible icons from making it to users 🎉)

### DIFF
--- a/packages/pharos/src/components/icon/pharos-icon.test.ts
+++ b/packages/pharos/src/components/icon/pharos-icon.test.ts
@@ -7,7 +7,9 @@ describe('pharos-icon', () => {
   let component: PharosIcon;
 
   beforeEach(async () => {
-    component = await fixture(html`<test-pharos-icon name="base"></test-pharos-icon>`);
+    component = await fixture(
+      html`<test-pharos-icon name="base" a11y-title="base-icon"></test-pharos-icon>`
+    );
   });
 
   it('is accessible', async () => {
@@ -15,9 +17,9 @@ describe('pharos-icon', () => {
   });
 
   it('throws an error for an invalid icon name', async () => {
-    component = await fixture(html` <test-pharos-icon name="fake"></test-pharos-icon> `).catch(
-      (e) => e
-    );
+    component = await fixture(
+      html` <test-pharos-icon name="fake" a11y-title="fake-icon"></test-pharos-icon> `
+    ).catch((e) => e);
     expect('Could not get icon named "fake"').to.be.thrown;
   });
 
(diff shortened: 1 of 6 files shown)


mtorres3 (approved):

packages/pharos/src/components/icon/pharos-icon.test.ts
```diff
@@ -74,11 +56,12 @@ describe('pharos-icon', () => {
     expect(title).to.contain.text(labelText);
   });
 
-  it('sets the svg title properly when description is set', async () => {
-    const labelText = 'This is a test title';
-    component.description = labelText;
-    await component.updateComplete;
-    const title = component.renderRoot.querySelector('svg>title');
-    expect(title).to.contain.text(labelText);
+  it('throws an erorr when neither a11y-title or a11y-hidden are set', async () => {
```
sirrah-tam, line 59:
```suggestion
  it('throws an error when neither a11y-title or a11y-hidden are set', async () => {
```

sirrah-tam (approved):

PR #695, opened 2024-03-02
Table: Create table components

### LINKED ISSUE #355: Tables: A way to organize information, especially large sets of information
Displaying structured information in a format of columns and rows can greatly increase the readability and scannability of that information for the end user.  Presenting information in this way also presents an opportunity for the reader to then have additional features like sorting and filtering to reduce the amount of information being presented so that it is more meaningful and even take an action if necessary for the given situation.

**Landscape analysis**

_Lists of content available_
The [JSTOR Browse pages](https://www.jstor.org/action/showJournals?contentType=journals&letter=I) are a way to list the content titles available on the platform.
Screen Shot 2022-06-01 at 5 24 16 PM

_Report requests_
JSTOR Administrators are able to request usage reports for making purchasing or other decisions based on how content is used, or not used.
Screen Shot 2022-06-01 at 5 08 42 PM

_Content curators and publishers_
Ingesting and publishing content metadata onto the JSTOR Platform.
Screen Shot 2022-06-01 at 5 10 17 PM

Another use case for content curators and publishers is understanding the success and failures when publishing content. Being able to quickly see what failed among potentially thousands of items so the user can correct and republish.
Screen Shot 2022-06-01 at 5 12 41 PM

Screen Shot 2022-06-01 at 5 15 39 PM

_Feature comparison/evaluation_
The [JSTOR Welcome Mat](https://www.jstor.org/artstor) for bringing Artstor content onto the JSTOR platform
Screen Shot 2022-06-01 at 5 18 21 PM

_Status of features_
Directly in the [Pharos site](https://pharos.jstor.org/components/component-status) there is a list of components and where they currently stand in development and release
Screen Shot 2022-06-01 at 5 21 52 PM

**Additional considerations**
Tables can get complex. There are a few cases located [on the WCAG site](https://www.w3.org/WAI/tutorials/tables/) like multiple heading levels, headings for both columns as well as the rows and captions/summaries to help explain the context of the table to a given assistive technology.

Dynamic display for multi-device support. Another potential consideration is around the ability to change the amount of columns/rows being displayed based on the viewable area of a given device. This allows for reducing/increasing the amount of information being presented to optimize the readability based off the viewable area the given user has.

Keeping context for large amounts of information. There are some cases where maintaining the headings of the table in the device viewport while the user is scrolling/navigating so as they are reading the information, they are quickly able to understand what a given column/row value is in reference.

Extensibility. Being able to layer features like sorting and filtering into a table empowers a user to customize the information they are presented so that it's more meaningful or take an action like making a purchase or correcting an error on a given task.

jialin-he:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Resolves #355 ; Create a basic table component

A table component that takes in user input data and display with pagination.


packages/pharos/src/components/table/pharos-table.scss
```diff
@@ -0,0 +1,35 @@
+:host {
+  display: inline-flex;
+  flex-direction: column;
+}
+
+.table {
+  border-collapse: collapse;
+
+  th {
+    border: 1px solid #eae8e1;
```
mtorres3, line 10: Is it possible to change the `#eae8e1` to `$pharos-color-marble-gray-base`?

packages/pharos/src/components/table/pharos-table.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding: var(--pharos-spacing-1-x);
+  }
+
+  td {
+    border: 1px solid #eae8e1;
```
mtorres3, line 15: Same here

mtorres3 (approved):

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,202 @@
...
+   *  },
+   *  {
+   *    name: Message
+   *    field: message
+   *  }
+   * ]
+   */
+  @property({ type: Array, reflect: true, attribute: 'columns' })
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  public columns: any[] = [];
```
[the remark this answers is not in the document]
jialin-he: Updated!

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import type { TemplateResult } from 'lit';
+
+import { html } from 'lit/static-html.js';
+
+import type { PharosTable } from './pharos-table';
+import type { PharosLink } from '../link/pharos-link';
+import type { PharosPagination } from '../pagination/pharos-pagination';
+import type { PharosSelect } from '../select/pharos-select';
+
+describe('pharos-table', () => {
```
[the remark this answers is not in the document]
jialin-he: Updated!

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    ) as PharosSelect;
+    selectDropdown['_select'].value = '2';
+    selectDropdown['_select'].dispatchEvent(new Event('change'));
+
+    await componentWithPagination.updateComplete;
+
+    pageNumber = componentWithPagination.renderRoot.querySelector(`.page-number-display`);
+    expect(pageNumber?.innerHTML).contains('1-');
+    expect(pageNumber?.innerHTML).contains('2 of');
+  });
```
daneah, line 96: I'm having trouble making sense of this test. Is there a reason we can't assert the entire set of data that's supposed to be displayed in the pagination?

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  it('is accessible with pagination', async () => {
+    await expect(componentWithPagination).to.be.accessible();
+  });
+
+  it('has 3 rows', async () => {
+    const rows = Array.prototype.slice.call(
+      component.renderRoot.querySelectorAll(`tr`)
+    ) as TemplateResult[];
+    expect(rows.length).to.be.eq(3);
+  });
```
daneah, line 71:
```suggestion
  it('has the correct number of rows', async () => {
    const rows = Array.prototype.slice.call(
      component.renderRoot.querySelectorAll(`tr`)
    ) as TemplateResult[];
    expect(rows.length).to.be.eq(3);
  });
```

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  it('render rows according to page size', async () => {
```
daneah, line 73:
```suggestion
  it('renders rows according to page size', async () => {
```

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    selectDropdown['_select'].dispatchEvent(new Event('change'));
+
+    await componentWithPagination.updateComplete;
+
+    pageNumber = componentWithPagination.renderRoot.querySelector(`.page-number-display`);
+    expect(pageNumber?.innerHTML).contains('1-');
+    expect(pageNumber?.innerHTML).contains('2 of');
+  });
+
+  it('update table after page size selection', async () => {
```
daneah, line 98:
```suggestion
  it('updates correctly after page size selection', async () => {
```

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,208 @@
...
+import { PharosSelect } from '../select/pharos-select';
+import { PharosPagination } from '../pagination/pharos-pagination';
+
+export type ColumnSpecification = {
+  name: string;
+  field: string;
+};
+
+export type RowData = {
+  [key: string]: string;
```
daneah, line 17: Can the row data values really only be strings? I could definitely imagine wanting to pass in `number`s or `null`s as well (maybe others?).

packages/pharos/src/components/table/pharos-table.test.ts
jialin-he: What do you mean by "entire set of data"? I was trying to verify that the `_pageStartNumber` and `_pageEndNumber` are calculated correctly. But they are private to the component so I have to look at the rendered html to verify, i.e the "(Displaying 1-2 of 2)" part.

packages/pharos/src/components/table/pharos-table.ts
jialin-he: updated it to "unknown", not sure if it is the best idea though....

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,208 @@
...
+  public hidePagination: boolean = true;
+
+  @property({ type: Number, reflect: true, attribute: 'total-results' })
+  public totalResults: number = 0;
+
+  @property({ type: Array, reflect: true, attribute: 'page-size-options' })
+  public pageSizeOptions: number[] = [50, 100];
+
+  @property({ type: String, reflect: true, attribute: 'caption' })
+  public caption: string = '';
```
kentjas1, line 81: I understand we need the caption for accessibility reasons, I'm curious if there should be any ability to visually hide it in certain cases or is the expectation should be having the practice of always showing a table caption?

brentswisher: ☝️ @sirrah-tam might have some insight into this, I think it could be "visually hidden" as long as it is still accessible to screen readers

packages/pharos/src/components/table/pharos-table.test.ts
daneah: That's kind of what I mean—why can't we test "Displaying 1-2 of 2" directly instead of these multiple, partial assertions?

packages/pharos/src/components/table/pharos-table.ts
sirrah-tam: It would be great to be able to include an attribute that auto-visually-hides the caption as that does appear to be pretty common in other examples I'm finding. Similar to how the <title> element gives a SVG it's accessible name (accNam), the <caption> gives the Table it's accName, helpful when users of AT are reviewing the contents of a page. Visually hiding the captions would mean that a sighted user would have to use other information on the page to understand the context and purpose of the table, ex page title or heading preceding the table.

packages/pharos/src/components/table/pharos-table.test.ts
jialin-he: Ah here is the reason! This is what I see after I test the whole sentence
image

packages/pharos/src/components/table/pharos-table.ts
jialin-he: If I understand correctly, we want to add another attribute, something like `auto-visually-hide = true`, to control if we want to visually hide the caption? Does that mean we should make the caption attribute a required attribute so that we always have a accessible name for the table?

packages/pharos/src/components/table/pharos-table.test.ts
daneah: Ah, perhaps you can try `innerText` instead of `innerHTML`?

packages/pharos/src/components/table/pharos-table.ts
kentjas1: I do feel like the caption should be required to enforce always having accessible tables generated but would leave that up to @sirrah-tam and the other wikkies to officially answer yes/no.

packages/pharos/src/components/table/pharos-table.test.ts
jialin-he: Yes it works!!!

packages/pharos/src/components/table/pharos-table.ts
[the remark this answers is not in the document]
jialin-he: I've been thinking about your suggestion Dane. I wonder if it is possible in some cases the content in a column can be in multiple types. Do we require the consumer to provide all possible types? Will that be too heavy load for the consumer?

packages/pharos/src/components/table/pharos-table.ts
jialin-he: Updated!

packages/pharos/src/components/table/pharos-table.ts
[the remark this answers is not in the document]
daneah: I would generally hope that if we're displaying tabular data, that within a column the data type is consistent (or can be coerced to be consistent by way of `string`, minimally). There's a tradeoff here between safety and tedium as you rightly call out...I don't really have strong feelings on this, and it can definitely be iterated on, so I don't want it to hold us up here too much!

packages/pharos/src/components/table/pharos-table.scss
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  .pagination {
+    height: var(--pharos-spacing-one-and-a-half-x);
+  }
+}
+
+.hide-table-caption {
+  visibility: hidden;
```
sirrah-tam, line 42: This CSS style will actually hide the content from a screen reader, similar to `display: none`. It's better to use the following set of styles when trying to visually hide content: 

```
.visually-hidden:not(:focus):not(:active) {
  clip: rect(0 0 0 0); 
  clip-path: inset(50%);
  height: 1px;
  overflow: hidden;
  position: absolute;
  white-space: nowrap; 
  width: 1px;
}
```

packages/pharos/src/components/table/pharos-table.ts
sirrah-tam: The caption element is not required per WCAG and there are some cases where the caption may not be needed, especially for simple tables where there is already a heading that precedes the table. However, I do like that we'd kind of take a stand that all tables need a caption by requiring it for our design system. I would however say we shouldn't have it hidden by default and instead make the user explicitly hide it.

packages/pharos/src/components/table/pharos-table.scss
brentswisher: I think this mixin is what you want within pharos https://github.com/ithaka/pharos/blob/e396a023873a9c5c949343919aee174570a9acd4/packages/pharos/src/utils/scss/_mixins.scss#L130

jialin-he: done!

daneah (approved):

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,221 @@
...
+   *    id: 2,
+   *    message: "second one"
+   *  }
+   * ]
+   */
+  @property({ type: Array, reflect: true, attribute: 'row-data' })
+  public rowData: RowData[] = [];
+
+  @property({ type: Boolean, reflect: true, attribute: 'hide-pagination' })
+  public hidePagination: boolean = true;
```
brentswisher, line 73: Might just be just personal opinion, but thoughts on making this `showPagination` instead and defaulting it to false? Setting hidePagination="false" to show the pagination seems like a more complicated mental model than just showPagination="true" or showPagination="false" ?

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        caption="test table with pagination"
+      >
+      </test-pharos-table>
+    `);
+  });
+
+  it('is accessible', async () => {
+    await expect(component).to.be.accessible();
+  });
+
```
brentswisher, line 67: Might be nice to add a unit test like [this one](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/button/pharos-button.test.ts#L185) which verifies it throws an error when there is no caption?

packages/pharos/src/components/table/pharos-table.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    await componentWithPagination.updateComplete;
+
+    rows = Array.prototype.slice.call(
+      componentWithPagination.renderRoot.querySelectorAll(`tr`)
+    ) as TemplateResult[];
+    expect(rows.length).to.be.eq(3);
+  });
+
+  it('fires a custom event pwhen go to prev and next page', async () => {
```
brentswisher, line 123:
```suggestion
  it('fires a custom event when going to previous and next page', async () => {
```

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,221 @@
...
+  @state()
+  private _pageSize = 50;
+
+  @state()
+  private _currentPage = 1;
+
+  protected override firstUpdated(): void {
+    this._pageSize = this.hidePagination ? this.rowData.length : this.pageSizeOptions[0];
+    this.totalResults =
+      this.hidePagination || !this.totalResults ? this.rowData.length : this.totalResults;
```
brentswisher, line 96: Why do we care about hidePagination here?

brentswisher (commented):
@jialin-he Just a couple of nits/questions, but this is looking awesome 🎉

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,221 @@
...
+  public totalResults: number = 0;
+
+  @property({ type: Array, reflect: true, attribute: 'page-size-options' })
+  public pageSizeOptions: number[] = [50, 100];
+
+  @property({ type: String, reflect: true, attribute: 'caption' })
+  public caption: string = '';
+
+  @property({ type: String, reflect: true, attribute: 'hide-caption-visually' })
+  public hideCaptionVisually: boolean = false;
```
sirrah-tam, line 85:
```suggestion
  public hideCaption: boolean = false;
```

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,221 @@
...
+  @property({ type: Number, reflect: true, attribute: 'total-results' })
+  public totalResults: number = 0;
+
+  @property({ type: Array, reflect: true, attribute: 'page-size-options' })
+  public pageSizeOptions: number[] = [50, 100];
+
+  @property({ type: String, reflect: true, attribute: 'caption' })
+  public caption: string = '';
+
+  @property({ type: String, reflect: true, attribute: 'hide-caption-visually' })
```
sirrah-tam, line 84:
```suggestion
  @property({ type: String, reflect: true, attribute: 'hide-caption' })
```

packages/pharos/src/components/table/pharos-table.ts
jialin-he: So if the table doesn't have a pagination, consumer doesn't have to provide a `totalResults` attribute, we can automatically calculate it from the rowData.

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -0,0 +1,221 @@
...
+        </div>`
+      : nothing;
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <table class="table">
+        <caption
+          class="${classMap({
+            [`visually-hidden`]: this.hideCaptionVisually,
```
sirrah-tam, line 204:
```suggestion
            [`visually-hidden`]: this.hideCaption,
```

sirrah-tam (commented):
Small request to change the reference to `hideCaption` vs `hideCaptionVisually` as more concise way of setting that attribute.

packages/pharos/src/components/table/pharos-table.test.ts
jialin-he: done!

packages/pharos/src/components/table/pharos-table.ts
jialin-he: Yeah I feel showPagination makes more sense, updated!

sirrah-tam (approved):

PR #737, opened 2024-03-28
Remove browser styled scrollbar from Tabs component

sirrah-tam:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

The update for the Tabs that allowed for horizontal scrolling causes a scroll bar to appear, even when there isn't a need for one. This added unnecessary padding just below the tabs to all instances of this pattern. These are styled slightly different in each browser as the browser applies its own styles. Below is an example from a Chromium browser.
Screenshot of Tab with scrollbar highlighted

Adding a style for the `.tab__list` element that sets the `scrollbar-width` to `none`.

Screenshot 2024-03-28 at 5 47 03 PM

**Additional context**
I confirmed that touch (mobile) users are still be able to tap and drag to scroll, mouse users will be required to scroll horizontally (either by the scroll wheel moving left/right or by holding down Shift and using the mouse scrollwheel) but won’t be able to click and drag, and keyboard-only users will still be able to tab through without hassle.

### DIFF
--- a/.changeset/eight-cows-bake.md
+++ b/.changeset/eight-cows-bake.md
@@ -1,0 +1,5 @@
+---
+"@ithaka/pharos": patch
+---
+
+Remove scrollbar for tabs
--- a/packages/pharos/src/components/tabs/pharos-tabs.scss
+++ b/packages/pharos/src/components/tabs/pharos-tabs.scss
@@ -31,6 +31,7 @@
   display: flex;
   overflow-x: scroll;
   scroll-behavior: smooth;
+  scrollbar-width: none; // removes the scrollbar browser style
   margin-bottom: var(--pharos-spacing-one-half-x);
   padding: var(--pharos-spacing-one-quarter-x) var(--pharos-spacing-one-quarter-x)
     var(--pharos-spacing-1-x);


daneah:
@sirrah-tam does this interfere with macOS users who have their system settings intentionally set to always show scroll bars?

mtorres3 (approved):

sirrah-tam:
> @sirrah-tam does this interfere with macOS users who have their system settings intentionally set to always show scroll bars?

It does in fact! That is a good catch and a TIL moment for me. After a bit more testing across browsers and OSs, it seems like we may not have a lot to do about this, and using the style as proposed would effectively hide the scroll bar for all users. 

It appears Safari on macOS doesn't show the scroll bar until its needed at that horizontal width, but both Chrome & Firefox will always show it when setting the system settings for scroll bars to "Automatically based on mouse or trackpad" and (obvs) "Always". When setting it to "When scrolling", the scroll bar does not appear unless the user scrolls

On Windows there is only the option to Always show the scroll bar or not, and not that 3rd option. Testing here finds that Firefox doesn't show it until it's necessary while Edge/Chrome always shows it regardless of horizontal width. 

Will close this PR unless a case is made that we remove the visual scroll bar altogether.

satya-achanta-venkata (approved):

PR #800, opened 2024-08-07
chore(sass): address deprecation warnings for nested rules

### LINKED ISSUE #783: Fix SCSS deprecation warnings
**The problem**

On the latest version of `sass`, we now get deprecation warnings about certain syntax that is quite present in the code base:

```
Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.
```

**The solution**

Update syntax to fix warnings. Either of the suggested approaches in the warning should be amenable in our code, but may need case-by-case consideration for which is more readable and so on. In very rare cases, changes may create a specificity issue—we should lean on Chromatic to let us know.

**Additional information**

[Sass docs](https://sass-lang.com/documentation/breaking-changes/mixed-decls/)

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Resolves #783 

Move declarations above nested rules


daneah:
Currently this attempt is causing styling issues, and trying the other approach of nesting things within `& { }` causes stylelint to conflict and fail. Revisiting later.

PR #813, opened 2024-09-03
Table: Add sticky header option

### LINKED ISSUE #802: Table: `hideCaption` property not visually hiding the caption
**Expected behavior**
When I set `hideCaption` to true, the table caption is visually hidden

**Actual behavior**
Even when `hideCaption` is set to true, the table caption is displayed

1. Create a table with a caption and set hideCaption to true
2. Load the page
3. See the caption

**Screenshots or code**
Screenshot 2024-08-29 at 10 58 10 AM

**Pharos version**
v14.4.0

**Your environment**

- OS: macOS
- Browser: Chrome
- Version: 127.0.6533.122

**Additional information**
It looks like the mapping to add the class name just isn't working correctly, as the class is rendering as blank spaces:

```html
<caption class="  "><!--?lit$5795360043$-->Table Caption</caption>
```

Adding a unit test to validate this functionality would probably be good to do too 😄

### LINKED ISSUE #805: Table: Add a sticky header option
**The problem**
If there is a lot of data in a table, it is hard to read. When I scroll down through the rows the table header scrolls out of view and it becomes harder to read the table because I have to remember what column means what. If it is a large table with horizontal scrolling, it becomes even more difficult to keep in mind what each column means.

**The solution**
There is an option added to the table component to make the header sticky.

**Additional information**
It looks like this may be possible by targeting the `<th>` elements directly with css: https://css-tricks.com/position-sticky-and-table-headers/

brentswisher:

- [x] Adds a new feature
- [x] Fixes a bug
- [x] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Closes #805 
Also closes #802 

Adds a `hasStickyHeader` property to the table component which sets `position:sticky` on the `thead` element and keeps it visible when the table is scrolled. 

Once the table is scrolled past the header (so the stickiness will display it over the table) a `table-header--has-active-sticky-header` class will toggle from an IntersectionObserver to set an elevation so it is more visually evident is it displaying over the content. 

**Additional context**
❤️  Shoutout to @daneah @elisa-vargas @jialin-he for working together on this at the working session last week!

Because it needed to use similar logic to conditionally set a class name as `hideCaption` we updated that logic to close #802 once we got it working. (classMap shouldn't be inside quotes)

I also updated the property annotation of the `hideCaption` to be a Boolean instead of a string (it is actually a boolean) - I don't think this should cause any problems, but want to call it out to verify my understanding of that.

### DIFF
--- a/packages/pharos/src/components/table/pharos-table.scss
+++ b/packages/pharos/src/components/table/pharos-table.scss
@@ -7,20 +7,25 @@
 
 .table {
   border-collapse: collapse;
-
-  th {
-    border: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
-    padding: var(--pharos-spacing-1-x);
-  }
-
-  td {
-    border: 1px solid var(--pharos-table-body-color-border, var(--pharos-color-ui-30));
-    padding: var(--pharos-spacing-1-x);
-  }
 }
 
 .table-header {
   background-color: var(--pharos-table-header-color-background, var(--pharos-color-ui-20));
+}
+
+.table-header--has-sticky-header {
+  position: sticky;
+  top: 0;
+}
+
+.table-header__cell {
+  border: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
+  padding: var(--pharos-spacing-1-x);
+}
+
+.table-body__cell {
+  border: 1px solid var(--pharos-table-body-color-border, var(--pharos-color-ui-30));
+  padding: var(--pharos-spacing-1-x);
 }
 
 .table-controls {
--- a/packages/pharos/src/components/table/pharos-table.scss
+++ b/packages/pharos/src/components/table/pharos-table.scss
@@ -18,27 +18,27 @@
   padding: var(--pharos-spacing-1-x);
 }
 
-.table-header--has-sticky-header {
+.table-body__cell {
+  border: 1px solid var(--pharos-table-body-color-border, var(--pharos-color-ui-30));
+  padding: var(--pharos-spacing-1-x);
+}
+
+.table-sticky-header {
   position: sticky;
   top: 0;
 }
 
-.table-header--has-active-sticky-header {
+.table-sticky-header--is-active {
   box-shadow: var(--pharos-elevation-level-2);
-
-  .table-header__cell {
-    border: none;
-    border-right: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
-
-    &:last-child {
-      border-right: none;
-    }
-  }
 }
 
-.table-body__cell {
-  border: 1px solid var(--pharos-table-body-color-border, var(--pharos-color-ui-30));
-  padding: var(--pharos-spacing-1-x);
+.table-sticky-header__cell--is-active {
+  border: none;
+  border-right: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
+
+  &:last-child {
+    border-right: none;
+  }
 }
 
 .table-controls {
--- a/packages/pharos/src/components/table/pharos-table.scss
+++ b/packages/pharos/src/components/table/pharos-table.scss
@@ -13,14 +13,27 @@
   background-color: var(--pharos-table-header-color-background, var(--pharos-color-ui-20));
 }
 
+.table-header__cell {
+  border: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
+  padding: var(--pharos-spacing-1-x);
+}
+
 .table-header--has-sticky-header {
   position: sticky;
   top: 0;
 }
 
-.table-header__cell {
-  border: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
-  padding: var(--pharos-spacing-1-x);
+.table-header--has-active-sticky-header {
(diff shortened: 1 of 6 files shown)


brentswisher:
Video of new sticky header interaction:

https://github.com/user-attachments/assets/8666abba-c8b3-4258-bed6-1593020b1f7d

packages/pharos/src/components/table/pharos-table.scss
```diff
@@ -7,20 +7,38 @@
...
+  box-shadow: var(--pharos-elevation-level-2);
+
+  .table-header__cell {
+    border: none;
+    border-right: 1px solid var(--pharos-table-header-color-border, var(--pharos-color-ui-30));
+
+    &:last-child {
+      border-right: none;
+    }
   }
```
daneah, line 36: Typically with BEM I think this would be `.table-header__cell--has-active-sticky-header` maybe? Involves adding a bit more fiddling at the render level, I suppose.

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -105,6 +118,41 @@ export class PharosTable extends ScopedRegistryMixin(PharosElement) {
     this.totalResults = !this.totalResults ? this.rowData.length : this.totalResults;
   }
 
+  private _initHeaderObserver(): void {
+    const ACTIVE_HEADER_CLASS = 'table-header--has-active-sticky-header';
+    if (!this.hasStickyHeader) {
+      throw new Error('Sticky header is not enabled, cannot initialize observer');
+    }
```
daneah, line 125: Where does this condition arise in practice? It feels overly defensive, but I might be missing something.

packages/pharos/src/components/table/pharos-table.scss
brentswisher: Yeah, I questioned that a bit too, no strong opinion really. I can try that out in the morning, we just need do a querySelectorAll for the `th` elements and loop over all of them to toggle a class with the IntersectionObserver flips.

I also wondered if we would be better served having separate `table-header` and `table-sticky-header` block level classes which would let the modifier be `--is-active` 

Having both `--has-sticky-header` and `--has-active-stick-header` feels weird to me, but I'm not particularly strong with BEM

packages/pharos/src/components/table/pharos-table.ts
brentswisher: Yeah, that's probably fair, got a little carried away, i can take it out

sirrah-tam (approved):
This looks good!

When we start adding functionality related to links/buttons/inline editing we should be aware of a couple of WCAG things. 

[2.4.11 Focus Not Obscured (Minimum) AA](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum)
[2.4.12 Focus Not Obscured (Enhanced) AAA](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced)

Re: 2.4.11 Focus Not Obscured (Minimum), since users are able to scroll the focused content if/when it becomes obscured we're in the clear. Only would be issue in the event the page loads while something is obscured and before the user takes any actions.

packages/pharos/src/components/table/pharos-table.scss
brentswisher: Okay, updated with block level classes for the sticky header, I updated the unit tests, I couldn't get the scrolling to work but it still has pretty good coverage

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -105,6 +118,52 @@ export class PharosTable extends ScopedRegistryMixin(PharosElement) {
     this.totalResults = !this.totalResults ? this.rowData.length : this.totalResults;
   }
 
+  private _toggleActiveStickyHeader = (active: boolean) => {
+    const ACTIVE_HEADER_CLASS = 'table-sticky-header--is-active';
+    const ACTIVE_HEADER_CELL_CLASS = 'table-sticky-header__cell--is-active';
```
shoupeva-ithaka, line 123: I am curious if you need an active cell class or if your sass can just style the children cells of the sticky header active class? Something like...
```sass
.table-sticky-header--is-active {
  ... styles
  th {
    ... styles
  }
}
```
... hard to know if this is possible as I cannot see the HTML in this PR. It is also more of nit, it's not like querying all the th elements is a very big deal.

packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -198,13 +257,18 @@ export class PharosTable extends ScopedRegistryMixin(PharosElement) {
     return html`
       <table class="table">
         <caption
-          class="${classMap({
-            [`visually-hidden`]: this.hideCaption,
-          })}"
+          class=${classMap({
+            ['visually-hidden']: this.hideCaption,
```
shoupeva-ithaka, line 261: Nice catch

shoupeva-ithaka (approved):

packages/pharos/src/components/table/pharos-table.ts
brentswisher: 👍 That was done in an [effort to be more BEM-like](https://github.com/ithaka/pharos/pull/813#discussion_r1742724972)

PR #825, opened 2024-10-25
Switch: New component contribution

### LINKED ISSUE #821: Toggle switch
Create a toggle switch component for a stylized way of showing enabled/disabled. 

**Landscape analysis**
Here is an example from AWS AppConfig
Image

https://shoelace.style/components/switch
https://m3.material.io/components/switch/overview
https://designsystem.porsche.com/v3/components/switch/examples
https://carbondesignsystem.com/components/toggle/usage/

**Additional considerations**
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role
https://www.w3.org/WAI/ARIA/apg/patterns/switch/

michael-iden:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [X] Changeset added?

Closes #821 by creating a new toggle switch component

Adds a stylized variant of a checkbox to indicate an enabled/disabled state and a mechanisms to update the state

**Additional context**


packages/pharos/src/components/switch/pharos-switch.scss
```diff
@@ -0,0 +1,127 @@
...
+.input-wrapper {
+  @include mixins.option-wrapper;
+
+  align-items: center;
+  column-gap: var(--pharos-spacing-1-x);
+}
+
+.switch__control {
+  height: 32px;
+  width: 70px;
```
daneah, line 12: Are these something that can/should be calculated from other values, or are they pretty squarely hand-crafted? Same question really for any hard-coded values, though some needing more scrutiny than others I'm sure.

packages/pharos/src/components/switch/pharos-switch.scss
```diff
@@ -0,0 +1,127 @@
...
+      right: var(--pharos-spacing-one-half-x);
+      background-color: var(--pharos-color-green-base);
+    }
+
+    &::after {
+      color: var(--pharos-color-green-base);
+      left: var(--pharos-spacing-one-half-x);
+      content: 'ON';
+    }
+  }
```
daneah, line 76: I wonder if these feel like an appropriate use of the `--pharos-color-feedback-success` alias token since it is sort of indicating feedback, though I understand it's not necessarily indicating "success"...

packages/pharos/src/components/switch/pharos-switch.wc.stories.jsx
daneah, line 1: Is it good to add an "Events" story like we have with some other components so folks can see the emitted events happening?

packages/pharos/src/components/switch/pharos-switch.ts
```diff
@@ -0,0 +1,110 @@
...
+          .checked=${this.checked}
+          role="switch"
+          ?disabled=${this.disabled}
+          @change=${this.onChange}
+        />
+        <div class="input-wrapper">
+          <label for="switch-element" class="switch__label" @click="${this._handleClick}">
+            <slot name="label"></slot>
+          </label>
+          <span class="switch__control" @click="${this._handleClick}"></span>
```
sirrah-tam, line 105: We may want to consider adding `aria-hidden="true"` to the span that visually makes up the switch control. Currently the CSS `content` is picked up by assistive tech (AT) when navigating w/ reading mode and adds to the verbosity of the component. Navigating through it in reading mode the user will discover the switch input first, then the visual label, and then the visual label for the state of the switch. Since the switch input already conveys the accessible name & state, this visual could be considered decorative and removed from AT navigation. This is similar to how we handle the Checkbox component ([3fd8e16](https://github.com/ithaka/pharos/commit/3fd8e16ed66e9c26bbbf901a114f2f50b563a1f2)) and the SVGs we use to visually represent the control.

packages/pharos/src/components/switch/pharos-switch.scss
michael-iden: From the mockup they appear to rather handcrafted. Technically 32px is 2rem but that could be mostly coincidental? @fswlee anything here you suggest or worth adding context to?

packages/pharos/src/components/switch/pharos-switch.scss
michael-iden: Hmm good question, here too. I think I'd be more apt to move this to success if we also made the light green background (green 97) a success token? Hypothetically could the success color become a greenish-blue, or whatever, and we'd also want to make sure the light variant of that gets updated as well?

packages/pharos/src/components/switch/pharos-switch.ts
michael-iden: Good call. I think I dropped this when moving it from the prior search/pals implementation on accident too.

michael-iden: added :D

packages/pharos/src/components/switch/pharos-switch.scss
fswlee: the little knob is 1rem x 1rem and there's .5rem space above ad below which creates the height. the width is a little more arbitrary. we could consider being more specific and intentional with our values, for sure. let me look at some options

packages/pharos/src/components/switch/pharos-switch.scss
daneah: Hmm a good point. Because I don't want to rush to that decision you outline there, I'm good with leaving these as-is for the moment!

packages/pharos/src/components/switch/pharos-switch.scss
michael-iden: Ahh i might have the knob 1px small in both dimensions right now at 15px. I can tweak as part of this too

packages/pharos-site/static/guidelines/switch.docs.tsx
```diff
@@ -0,0 +1,172 @@
...
+        storyBookType="forms"
+      >
+        <PharosSwitch>
+          <span slot="label">I am a switch</span>
+        </PharosSwitch>
+      </PageSection>
+      <PageSection
+        topMargin
+        title="Usage"
+        description="The switch component is used to toggle if an control is enabled or disabled. The value of the switch is persisted at the moment the control is updated, without a separate save mechanism."
```
brentswisher, line 23:
```suggestion
        description="The switch component is used to toggle if a control is enabled or disabled. The value of the switch is persisted at the moment the control is updated, without a separate save mechanism."
```

packages/pharos/src/components/switch/pharos-switch.scss
```diff
@@ -0,0 +1,122 @@
...
+  display: flex;
+  align-items: center;
+  box-sizing: border-box;
+  cursor: pointer;
+  border: 1px solid;
+  border-color: var(--pharos-color-marble-gray-80);
+  border-radius: 2rem;
+  background-color: var(--pharos-color-marble-gray-94);
+  transition:
+    background-color 0.3s ease-out,
```
brentswisher, line 22: I do like the way this works, but see it is not one of our [transition tokens](https://pharos.jstor.org/design-tokens/transitions). I wonder if switching to `--pharos-transition-duration-default` which is 250 ms so it matched or other transitions would be make a noticeable difference?(cc @fswlee as this is mostly a design decision)

brentswisher:
🤔 Running pharos site locally, I am not seeing the component render in the examples. but I am seeing the label. 

Screenshot 2024-10-30 at 8 33 40 AM

brentswisher:
> 🤔 Running pharos site locally, I am not seeing the component render in the examples. but I am seeing the label.
 Ah, never mind, the site just needs the new component added where it is initializing Pharos components, I can push a fix for that

brentswisher (approved):
Left one note about the transition length, but I'm good with whatever way that decisions goes, its looking really good!

sirrah-tam (approved):
This looks and tests great! Verified against VoiceOver + Safari/Chrome on MacOS and JAWS + Edge/Chrome & NVDA + Firefox on Windows.

packages/pharos/src/components/switch/pharos-switch.scss
daneah: FWIW 50ms is generally imperceptible to humans so I'd go for the token value too 😄

michael-iden: ✅

Issue #826, opened 2024-10-28
Modal: Y Scrollbar region always present when using a mouse


michael-iden:
**Expected behavior**
Scrollbar region only is shown only when necessary based off the contents of the modal overflowing 

**Actual behavior**
Y Scrollbar region always present when a mouse input device is present 

**Steps to reproduce the issue**
1. Plug in a mouse
2. Open a modal
3. See that the modal doesn't scroll but the scroll region is still visible

**Screenshots or code**
![Image](https://github.com/user-attachments/assets/9380f01f-5785-455e-b6a5-6017215ca32e)

**Pharos version**
14.4.0

**Your environment**

- OS: macOS
- Browser: Chrome, Firefox, Safari
- Version: Latest for each as of 2024-10-28

**Additional information**
https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/modal/pharos-modal.scss#L98

brentswisher:
🤔 This doesn't happen in the [storybook examples](https://pharos.jstor.org/storybook/?path=/story/web-components_components-modal--events) that I can see, I wonder if something on the css of that page it making it think that the content is taller than the modal even though it doesn't look like it

brentswisher:
🤔 The plot thickens, I don't see it even on that page you shared in the screenshot:
![Image](https://github.com/user-attachments/assets/e95fe72a-b93a-4f16-b256-f04921865f7a)

michael-iden:
@brentswisher do you have a mouse plugged in? There are some mac level settings that can also affect this

![Image](https://github.com/user-attachments/assets/b23aeaaa-68eb-4e09-9aaa-7cfefad5494c)

brentswisher:
Hmm, interesting. I do have a Magic Mouse connected via Bluetooth

brentswisher:
I vaguely remembered a similar conversation and it looks like this maybe came up once before in #531 but was deemed out of scope - not that it shouldn't be looked into, but just as additional info.

Are you using a wired mouse?

daneah:
@michael-iden @brentswisher as mentioned this has been raised in the past (offline as well) and we have closed citing the macOS-level settings as the fix. I believe that if a user has a mouse plugged in such that the Pharos modal shows this scroll bar, they would have a similar experience regularly on other sites across the web as well.

michael-iden:
[...]

PR #827, opened 2024-10-28
fix(modal & tabs): remove overflow property that causes unnecessary scrollbars and update to Node 22

michael-iden:

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [X] Changeset added?

#826 

Remove overflow property for an element that will never overflow

**Additional context**


.tool-versions
```diff
@@ -1 +1 @@
-nodejs 20.11.1
+nodejs 20.18.0
```
daneah, line 1: I recommend we move to Node 22 if possible [since that's what the docs say we should be using](https://github.com/ithaka/pharos/blob/develop/docs/development/quick-start.md?plain=1#L21), and that we update the docs, maybe to remove reference to a version at all.

michael-iden: Interesting. I was curious and it looks like [node 22 LTS support starts tomorrow](https://github.com/nodejs/Release). I can bump that and clean up the docs
image

docs/development/quick-start.md
```diff
@@ -18,7 +18,7 @@ Start by cloning the Pharos repository:
 $ git clone git@github.com:ithaka/pharos.git
 ```
 
-Pharos supports the current long-term support (LTS) version of Node.js—currently version 20. Node can be installed from [its website](https://nodejs.org/en/), with [nvm](https://github.com/nvm-sh/nvm#install--update-script), or with [asdf](https://asdf-vm.com/) and its [Node.js plugin](https://github.com/asdf-vm/asdf-nodejs).
+Refer to [.tool-versions](../../.tool-versions) for the latest node version compatible with Pharos. Node can be installed from [its website](https://nodejs.org/en/), with [nvm](https://github.com/nvm-sh/nvm#install--update-script), or with [asdf](https://asdf-vm.com/) and its [Node.js plugin](https://github.com/asdf-vm/asdf-nodejs).
```
brentswisher, line 21: I feel like we lose a little bit of accountability by removing the statement that "Pharos supports the current LTS release of Node.js" - It seems partially a statement of intent?

brentswisher:
I see the [size report](https://github.com/ithaka/pharos/actions/runs/11893249906/job/33326654553?pr=827) action is failing, wondering if that is because of the node 22 update?

docs/development/quick-start.md
daneah: Yeah I would like to keep the statement of intent and remove only the specific version.

michael-iden: Could we add renovate and a statement about using that to stay on top of dependencies to recover accountability while also being more accurate about what we currently support as this change intends to do?

michael-iden:
@brentswisher I also noticed the size job failing and it does relate to node 22 but I don't know how to fix it. It _might_ be a task that relies the latest build. 

My interpretation of the error is that the job runs a build against both the current branch and develop with node22, and its failing to run against develop because it has the unsupported node 20 syntax. So it should get corrected once we merge and develop gets these updates to evaluate in the next PR

packages/pharos/src/components/tabs/pharos-tabs.scss
```diff
@@ -29,7 +29,7 @@
 
 .tab__list {
   display: flex;
-  overflow-x: scroll;
+  overflow-x: auto;
```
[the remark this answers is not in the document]
michael-iden: The intent here is a little different between the modal and the tabs. 
- For the tabs we do want the scrolling behavior and the scroll bar to show up when we need it, which we gain by using `auto`
- With the modal, we never want the region to scroll (at least with the current understanding of the UX, [your prior point from issue](https://github.com/ithaka/pharos/issues/826#issuecomment-2441923650) the still has plenty of merit)

docs/development/quick-start.md
brentswisher: I'm all for adding Renovate, but that feels like a separate thing a bit? Since this is the quickstart, it's nice to be able to tell people just getting started we use the latest LTS version as our baseline.

What about something like:
> "Pharos is committed to supporting the current long-term support (LTS) version of Node.js. Refer to the [.tool-versions](../../.tool-versions) file for the latest version compatible with Pharos."

michael-iden: Good suggestion. Thanks!

daneah: FWIW there is an open issue for using Dependabot for automated dependency updates, which I would currently still be inclined to use so that Pharos is not using proprietary infrastructure. But we should discuss it more there.

brentswisher (approved):

PR #891, opened 2025-02-24
Combobox: allow disabled options

### LINKED ISSUE #890: Add ability to disable options in a combobox
**The problem**
When using a combobox to allow the selection of something from a large list of items, there is no way to show the user an item, but also denote it is not available for selection. 

**The solution**
Allow options to be added to a combobox with a disabled attribute, which will show in search results but not be selectable in the combobox

**Alternatives considered**
One alternative would be to have items not available for selected be removed from the list entirely, but that can be confusing, If someone is searching for something they know exists, and it is not showing up, it seems more likely the search is broken or the list is incomplete, rather than the item is found but not available for selection.

brentswisher:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [X] Changeset added?

Closes #890 

Adds the ability to include disabled items in a combobox, so they appear in the the list but are not selectable.

If an option is passed into a combobox, when the list of options is rendered, it will appear in the list but be in a disabled state.

**Screenshot:**
Screenshot 2025-02-24 at 2 33 38 PM

**Additional context**
I also updated the React storybook to have a bunch of states that were missing compared to the Web Components version

Accessibility is a little tricky here, but I _think_ I have it right - https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
- The items appear in the list, but have `aria-disabled="true"` 
- The combobox maintains focus while options are navigated through
- If an option is disabled, it **is not** navigable with the arrow keys
- When using a screen ready, If an option is disabled, it **is navigable** when using the reader "Modifier" (Ctrl+option in VoiveOver) +  arrow keys, and they are announced as "dimmed"
- When clicking or pressing enter on a disabled option, the value is not updated

https://github.com/user-attachments/assets/760f6c8f-1b54-4769-9597-7188c1ff02d4

### DIFF
--- a/packages/pharos/src/components/combobox/pharos-combobox.scss
+++ b/packages/pharos/src/components/combobox/pharos-combobox.scss
@@ -185,8 +185,7 @@
   background-color: var(--pharos-dropdown-menu-item-color-background-disabled);
   cursor: default;
 
-  &:hover,
-  &[highlighted] {
+  &:hover {
     color: var(--pharos-dropdown-menu-item-color-text-disabled);
     background-color: var(--pharos-dropdown-menu-item-color-background-disabled);
     cursor: default;


packages/pharos/src/components/combobox/PharosCombobox.react.stories.jsx
```diff
@@ -58,39 +58,59 @@ export const Base = {
...
         <option value="1">Option 1</option>
         <option value="2">Option 2</option>
         <option value="3">Option 3</option>
       </PharosCombobox>
     </div>
   ),
-  args: {
-    placeholder: 'Enter some text',
-  },
+  args: { placeholder: 'Enter some text' },
```
daneah, line 108: It seems like you could delete this now as I don't see any references to `args.placeholder` remaining.

packages/pharos/src/components/combobox/pharos-combobox.scss
```diff
@@ -180,6 +180,19 @@
   @include selected-text;
 }
 
+.combobox__option--disabled {
+  color: var(--pharos-dropdown-menu-item-color-text-disabled);
+  background-color: var(--pharos-dropdown-menu-item-color-background-disabled);
+  cursor: default;
+
+  &:hover,
+  &[highlighted] {
```
daneah, line 189: Not sure I understand the presence of this `[highlighted]` rule; can you elaborate?

brentswisher: Ah, you are right, it can be removed. In earlier versions of this, it was possible to navigate to the disabled options and this was added to fix an issue making it no longer look disabled due to specificity issues. But that should never happen now

jialin-he (approved):

daneah (approved):

PR #897, opened 2025-03-03
feat(progress-bar): Add  indeterminate progress bar variant

### LINKED ISSUE #878: Create indeterminate progress bar
Create an indeterminate progress bar that shows we are processing the operation in the background without giving the user an estimate on when the operation will finish. 

**Landscape analysis**
https://carbondesignsystem.com/components/progress-bar/usage/#indeterminate-progress-bar

**Additional considerations**
Does this fall more under progress bar or loading spinner?

ymouzakis:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

Resolves #878 

Make the `value` property optional, rendering an indeterminate progress bar if unspecified.

### DIFF
--- a/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
+++ b/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
@@ -19,6 +19,12 @@
   transition: width 0.6s ease;
 }
 
+.progress-bar--indeterminate {
+  background-color: var(--pharos-color-glacier-blue-base);
+  animation: indeterminateAnimation 1s infinite linear;
+  transform-origin: 0% 50%;
+}
+
 .progress-bar__title {
   @include mixins.font-base(
     $font-size: var(--pharos-type-scale-4),
@@ -32,19 +38,11 @@
   color: var(--pharos-color-text-40);
 }
 
-.progress-bar--indeterminate {
-  width: 100%;
-  height: 100%;
-  background-color: var(--pharos-color-glacier-blue-base);
-  animation: indeterminateAnimation 1s infinite linear;
-  transform-origin: 0% 50%;
-}
-
 @keyframes indeterminateAnimation {
   0% {
     transform:  translateX(0) scaleX(0);
   }
-  
+
   40% {
     transform:  translateX(0) scaleX(0.4);
   }
--- a/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
+++ b/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
@@ -44,9 +44,11 @@
   0% {
     transform:  translateX(0) scaleX(0);
   }
+  
   40% {
     transform:  translateX(0) scaleX(0.4);
   }
+
   100% {
     transform:  translateX(100%) scaleX(0.5);
   }
--- a/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
+++ b/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
@@ -31,3 +31,23 @@
 
   color: var(--pharos-color-text-40);
 }
+
+.progress-bar-indeterminate {
+  width: 100%;
+  height: 100%;
+  background-color: var(--pharos-color-glacier-blue-base);
+  animation: indeterminateAnimation 1s infinite linear;
+  transform-origin: 0% 50%;
+}
+
+@keyframes indeterminateAnimation {
+  0% {
+    transform:  translateX(0) scaleX(0);
+  }
+  40% {
+    transform:  translateX(0) scaleX(0.4);
+  }
+  100% {
+    transform:  translateX(100%) scaleX(0.5);
+  }
+}
--- a/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
+++ b/packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
@@ -32,7 +32,7 @@
   color: var(--pharos-color-text-40);
 }
 
-.progress-bar-indeterminate {
+.progress-bar--indeterminate {
   width: 100%;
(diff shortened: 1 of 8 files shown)


shoupeva-ithaka (approved):

packages/pharos/src/components/progress-bar/pharos-progress-bar.scss
```diff
@@ -31,3 +31,25 @@
 
   color: var(--pharos-color-text-40);
 }
+
+.progress-bar-indeterminate {
```
daneah, line 35: Nit: `indeterminate` feels like a modifier of progress bar, so using `--` better reflects that in BEM-style CSS (needs adjustment in the component code as well, if changing):

```suggestion
.progress-bar--indeterminate {
```

packages/pharos/src/components/progress-bar/storyArgs.ts
```diff
@@ -4,6 +4,13 @@ export const defaultArgs = {
   description: 'Processing',
 };
 
+export const indeterminateArgs = {
+  title: 'Indeterminate',
+  variant: 'indeterminate',
+  value: 10,
```
daneah, line 10: I'm sort of wondering if `indeterminate` is really a variant of the progress bar, or if it's more like it has an `undefined` value...

ymouzakis: Oh so like if there's no value, it defaults to being indeterminate? If that's what you mean, IDK it seems like a very non-deliberate way to set that experience imo. I also think that it might cause some unforeseen issues as progress values are being set.

packages/pharos/src/components/progress-bar/pharos-progress-bar.ts
```diff
@@ -39,15 +62,8 @@ export class PharosProgressBar extends PharosElement {
         class="progress-bar__wrapper"
```
sirrah-tam, line 62: Would it be possible to add the `aria-describedby` attribute on the role=progressbar that points to a new ID on the `div.progress-bar__description`? This will allow the progressbar to have that extra context all within the component itself, rather than requiring an AT user to navigate forward to discover the associated text.

packages/pharos/src/components/progress-bar/storyArgs.ts
daneah: I think we may disagree here; having the default behavior be an indeterminate bar if I haven't specified a value sort of feels expected to me! What kinds of unforeseen issues are you thinking about (I guess if you have any in mind, they're foreseen 😆)?

daneah: I think that can be managed through the component life cycle, if that's the DX we want for consumers. But I also don't have terribly strong feelings about it 😄 would welcome additional opinions one way or the other, in that case.

brentswisher: 🤔 I think if I'm in favor of the variant - though I do see both sides. If I wanted to display an indeterminate bar, I would expect to have to opt into it. 

For a bar that will never switch to the determinate version, It also seems like this would be easier to understand if I stumble upon it:
```
<storybook-pharos-progress-bar variant="indeterminate">
  ```
as opposed to :
```
<storybook-pharos-progress-bar value="undefined">
```

**Comment by brentswisher** [comment]
@ymouzakis Could you update the React Storybook story too?

File: packages/pharos/src/components/progress-bar/storyArgs.ts (hunk 2 above)
Comment (daneah) [reply packages/pharos/src/components/progress-bar/storyArgs.ts:10]: Well, to be fair, what I'm suggesting is that it would just be `<storybook-pharos-progress-bar>` without any property.

Comment (brentswisher) [reply packages/pharos/src/components/progress-bar/storyArgs.ts:10]: 🤔 Yeah that's fair

File: packages/pharos/src/components/progress-bar/pharos-progress-bar.ts [hunk 4]
```diff
@@ -36,18 +59,12 @@ export class PharosProgressBar extends PharosElement {
...
-              ${PharosColorGlacierBlueBase} ${this.value < 35 ? '100%' : 135 - this.value + '%'},
-              ${PharosColorNightBlueBase}
-            )`,
-            width: `${this.value}%`,
-          })}
+          class="${classList}"
+          style=${progressStyle}
         ></div>
       </div>
       <div class="progress-bar__description">
```
sirrah-tam, line 70:
```suggestion
      <div class="progress-bar__description" id="description">
```

Ensuring that the newly added `aria-describedby` has a valid ID ref.

packages/pharos/src/components/progress-bar/storyArgs.ts
ymouzakis: @daneah ok, I am convinced. I was thinking of it the same way as Brent with value="undefined" Thanks for the discussion ya'll! 😄

ymouzakis: Would this technically be a breaking change since the default version of the progress bar is different now?

daneah: This is a common point of contention—technically the API hasn't changed, but you're right that the outcome has certainly changed. I would advocate that here we call it a minor change since right now the behavior when nothing is specified feels ill-defined.

daneah: Again, I think if we manage the initial rendering through life cycling we should be able to avoid such a blip?

daneah (approved):

PR #904, opened 2025-03-11
Add catalog-info.yaml

eskimoquinn:
This PR adds a `catalog-info.yaml` file to Backstage when merged. If your repo is GitHub Only, this will not be picked up Backstage presently.


daneah:
Thanks @eskimoquinn. Because this is an open source project, this is already handled in another private repository.

PR #915, opened 2025-03-26
Table: Support custom header

adandavati:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

This updates the pharos table to be able to support any kind of elements in column headers. It also removed the `ResizeObserver` in the table cells which introduced a performance issue on even relatively large tables.

Adds new custom elements that can be used like the <thead> and <th> elements to create headers with any kind of content inside them

**Additional context**
Removing the ResizeObserver does re-introduce the bug it was fixing (table cell slots don't take the full height of the table cells) but that seems a good tradeoff vs the current experience of the page breaking for large tables.


packages/pharos/src/components/table/pharos-table.ts
```diff
@@ -122,39 +140,27 @@ export class PharosTable extends ScopedRegistryMixin(PharosElement) {
...
         entries.forEach((entry) => {
-          this._toggleActiveStickyHeader(entry.isIntersecting);
+          this._toggleElevatedStickyHeader(entry.isIntersecting);
         });
       },
       {
         root: this.shadowRoot?.querySelector('.table'),
         // Negative top rootMargin to offset the viewbox used for intersection calculations by the height of the current header
-        rootMargin: `-${this.header.getBoundingClientRect().height}px 0px 0px 0px`,
+        rootMargin: `-${height}px 0px 0px 0px`,
```
brentswisher, line 163: With slotted content, the header doesn't have a height until all its child elements have rendered, so we need to pass it in to be sure it isn't 0

brentswisher (approved):

daneah (approved):

PR #916, opened 2025-03-27
fix(switch): make checked a boolean attribute

michael-iden:

- [ ] Adds a new feature
- [X] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Occasionally and inexplicably, the pharos switch would be marked as checked when the condition to mark it as such was `false`. My best guess is that pieces at play (browser, vue, pharos, lit, native checkbox) didn't realize we were working with a checkbox and somewhere along the line the checked attribute was treated as a regular attribute instead of a boolean attribute. 

Make `checked` a boolean attribute instead of a regular attribute in lit

**Additional context**
https://lit.dev/docs/v2/templates/expressions/#boolean-attribute-expressions


daneah (approved):

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -55,12 +55,14 @@ export class PharosCheckbox extends FormMixin(FormElement) {
   @query('#checkbox-element')
   private _checkbox!: HTMLInputElement;
 
+  private _defaultChecked = false;
```
michael-iden, line 58: I was having a hard time getting the `_defaultChecked` to work when using the boolean attributes for both the checkbox and switch components. It seemed the reset wasn't working properly or reset was resetting too much based off the output when running the failing unit tests. 

I saw that `defaultChecked` was an attribute of the [HTMLInputElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultChecked) but by extracting the state to the component, I had much better behavior in the unit tests.

packages/pharos/src/components/switch/pharos-switch.ts
```diff
@@ -51,6 +53,7 @@ export class PharosSwitch extends FormMixin(FormElement) {
     const originalCheckedState = !this._switch.checked;
 
     this.checked = this._switch.checked;
+    console.log('onChange', this.checked);
```
daneah:
```suggestion
```

daneah (changes requested):

daneah (approved):

PR #943, opened 2025-04-25
chore(deps): upgrade to Lit 3 and latest web component deps

### LINKED ISSUE #634: Upgrade to Lit 3
**The problem**

[Lit 3 was recently released](https://lit.dev/blog/2023-10-10-lit-3.0/). We're on Lit 2 and don't want to fall too far behind. Lit 3 claims to largely be a smooth upgrade, as one of the only known breaking changes is dropping IE 11 support (which we already do not support).

**The solution**

Upgrade to Lit 3.

**Additional information**

I tried briefly installing Lit 3 and running a build, but ran into a fair number of TypeScript errors as a result. These could be because Lit 3 introduced breaking changes they didn't realize.

### LINKED ISSUE #745: Get on latest scoped web component-related dependencies
**The problem**
I want the project to be on the latest and greatest of as many dependencies as possible to ensure we stay apace of the ecosystem.

**The solution**
* Upgrade `@lit-labs/scoped-registry-mixin`
* Upgrade `@webcomponents/scoped-custom-element-registry`

**Additional information**
* `@lit-labs/scoped-registry-mixin` is blocked because on `1.0.2`+ it pulls in Lit 3 (#634)
* `@webcomponents/scoped-custom-element-registry` is blocked because on `0.0.4`+ something causes some unit tests to start failing ([changelog](https://github.com/webcomponents/polyfills/blob/master/packages/scoped-custom-element-registry/CHANGELOG.md#004---2022-01-27)). NOTE: Although the dependency is listed in `pharos-site`'s dependencies, it is `require`d during testing in the test config.

### LINKED ISSUE #866: Site: @babel/core update breaks Gatsby build
**Expected behavior**
Updating a patch version of `@babel/core` should be okay.

**Actual behavior**
Updating from 7.26.0 to 7.26.7 breaks the Gatsby site build.

1. run e.g. `yarn upgrade-interactive` and bump `@babel/core`
2. run `yarn site:build`
3. Observe an error regarding use of `document` in `@ithaka/focus-trap` that isn't available during SSR

**Additional information**
It's possible that resolving #777 will also fix this.

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Resolves #634 
Resolves #745 
Resolves #866

- Bump Lit to v3 and bump custom elements dependencies to latest
- Remove quotes from attributes and props in templates, because they’re unnecessary and it better calls to attention the static vs. dynamic values, and further reduces inconsistency/“moving parts” around how they’re handled
- Use `.` instead of `?` in stories, because Storybook does not seem to handle boolean syntax in Lit correctly at the end of the day.
- Remove unnecessary `="true"` from some boolean attributes since that’s implied


daneah:
There remain two tricky issues here with properties not updating as part of the Lit render cycle, and I can't really figure them out.

The first is the `icon-condensed` property of the Button. One thing that could be notable is that `icon-condensed` is only ever used in the Button's styles, so it might be that Lit is not subscribing to changes since it isn't in the render cycle. What's much weirder is that Button also has a `large` property, which works pretty identically except that it increases rather than decreases the `padding` of the Button. I don't see how `icon-condensed` and `large` are any different from one another, other than that `icon-condensed` has the dash. I tried altering this to just `condensed` to see if it made a difference, to no avail.

The second is the `full-width` property of the Tooltip. This is used in calculating the size and placement of the tooltip, which is redone each time the tooltip's `open` property changes.

I note that in both cases in Storybook, the attributes aren't even being set on the host elements in the DOM like they should be. If you manually render e.g. `icon-condensed` into the DOM for the story, things work, but it remains non-reactive because it isn't being tied to `args`.

I might call on you @Niznikr in case this is a familiar tale at all?

Niznikr:
Are these issues only happening in SB? If so, It might be worth [exploring this tool](https://github.com/break-stuff/wc-storybook-helpers), particularly [this feature](https://github.com/break-stuff/wc-storybook-helpers?tab=readme-ov-file#render-default-attribute-values).

daneah:
@Niznikr yeah it's happening in Storybook, and my worry is that because it broke during this migration, something changed about the actual reactivity life cycle for our components, so trying to gain some confidence that it's just Storybook and Lit not playing nice with each other. I will have a look at `wc-storybook-helpers` 😄

daneah:
Now that we're on Storybook v9 I no longer see the last remaining issues I mentioned above. Those appear to have been issues with Storybook more than issues with Lit and the components in practice. I think this is ready to move on, pending fixing up this issue with the Netlify build for the Gatsby site.

packages/pharos/src/components/button/pharos-button.ts
```diff
@@ -299,8 +299,8 @@ export class PharosButton extends ScopedRegistryMixin(FocusMixin(AnchorElement))
             id="button-element"
             name="${ifDefined(this.name)}"
             value="${ifDefined(this.value)}"
-            ?autofocus=${this.autofocus}
-            ?disabled=${this.a11yDisabled ? false : this.disabled}
+            .autofocus=${this.autofocus}
+            .disabled=${this.a11yDisabled ? false : this.disabled}
```
michael-iden: I'm curious on the change to move from [boolean attributes](https://lit.dev/docs/templates/expressions/#boolean-attribute-expressions) to [property expressions](https://lit.dev/docs/templates/expressions/#property-expressions)

From the docs I read 
> You can use the property expression syntax to pass complex data down the tree to subcomponents.

which might not apply so much for these attributes are boolean attributes on the underlying HTML elements?

daneah: In short, these weren't working in Storybook without doing this. Last night I tried a fresh branch that upgraded the dependencies without making these changes, and Storybook was still broken. I would prefer not to change these, as I think my mental model matches yours, but I have been so far unable to figure out what's wrong. I'd be happy to pair on that and show you what I'm seeing at some point!

daneah: I reverted the changes in the components themselves, leaving only the changes in the stories templates. I did make a pass to convert all `some-attr="${this.someThing}"` to be `some-attr=${this.someThing}` for consistency and because quotes aren't necessary, and it also better visually distinguishes static string values from calculated ones. I don't think this should have any functional effect in practice, but reduces moving parts on this somewhat open question.

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -140,11 +140,11 @@ export class PharosCheckbox extends FormMixin(FormElement) {
         .value=${this.value}
         .indeterminate=${this.indeterminate}
         ?checked=${this.checked}
```
brentswisher, line 142:
```suggestion
        .checked=${this.checked}
```
Update here too?

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -140,11 +140,11 @@ export class PharosCheckbox extends FormMixin(FormElement) {
...
-        ?required="${this.required}"
+        ?required=${this.required}
```
brentswisher, line 143:
```suggestion
        .required=${this.required}
```

packages/pharos/src/components/checkbox/pharos-checkbox.ts
```diff
@@ -140,11 +140,11 @@ export class PharosCheckbox extends FormMixin(FormElement) {
...
         ?disabled=${this.disabled}
```
brentswisher, line 144:
```suggestion
        .disabled=${this.disabled}
```

packages/pharos/src/components/combobox/pharos-combobox.ts
```diff
@@ -469,20 +469,20 @@ export class PharosCombobox extends ScopedRegistryMixin(FormMixin(FormElement))
...
-          name="${this.name}"
+          name=${this.name}
           type="text"
-          .value="${this._displayValue}"
-          ?required="${this.required}"
-          ?disabled="${this.disabled}"
-          placeholder="${this.placeholder}"
+          .value=${this._displayValue}
+          ?required=${this.required}
+          ?disabled=${this.disabled}
```
brentswisher, line 476:
```suggestion
          .required=${this.required}
          .disabled=${this.disabled}
```

packages/pharos/src/components/radio-button/pharos-radio-button.ts
```diff
@@ -83,10 +83,10 @@ export class PharosRadioButton extends FormMixin(FormElement) {
         type="radio"
         .value=${this.value}
         .checked=${this.checked}
-        ?required="${this.required}"
+        ?required=${this.required}
```
brentswisher, line 86:
```suggestion
        .required=${this.required}
```

packages/pharos/src/components/radio-button/pharos-radio-button.ts
```diff
@@ -83,10 +83,10 @@ export class PharosRadioButton extends FormMixin(FormElement) {
...
         ?disabled=${this.disabled}
```
brentswisher, line 87:
```suggestion
        .disabled=${this.disabled}
```

packages/pharos/src/components/select/pharos-select.ts
```diff
@@ -98,11 +98,11 @@ export class PharosSelect extends ScopedRegistryMixin(
           id="select-element"
           name=${this.name}
           .value=${this.value}
-          ?required="${this.required}"
+          ?required=${this.required}
           ?disabled=${this.disabled}
```
brentswisher, line 102:
```suggestion
          .required=${this.required}
          .disabled=${this.disabled}
```

packages/pharos/src/components/text-input/pharos-text-input.ts
```diff
@@ -256,20 +256,20 @@ export class PharosTextInput extends ScopedRegistryMixin(FormMixin(FormElement))
...
           id="input-element"
-          name="${this.name}"
-          .type="${this.type}"
+          name=${this.name}
+          .type=${this.type}
           .value=${this.value}
-          ?required="${this.required}"
+          ?required=${this.required}
           ?disabled=${this.disabled}
           ?readonly=${this.readonly}
```
brentswisher, line 264:
```suggestion
          .required=${this.required}
          .disabled=${this.disabled}
          .readonly=${this.readonly}
```

packages/pharos/src/components/textarea/pharos-textarea.ts
```diff
@@ -173,18 +173,18 @@ export class PharosTextarea extends FormMixin(FormElement) {
           cols=${ifDefined(this.cols)}
           rows=${ifDefined(this.rows)}
           dirname=${ifDefined(this.dirname)}
-          name="${this.name}"
+          name=${this.name}
           .value=${this.value}
-          ?required="${this.required}"
+          ?required=${this.required}
           ?disabled=${this.disabled}
           ?readonly=${this.readonly}
```
brentswisher, line 180:
```suggestion
          .required=${this.required}
          .disabled=${this.disabled}
          .readonly=${this.readonly}
```

brentswisher (approved):
🎉

PR #969, opened 2025-06-03
chore(storybook): upgrade to Storybook 9

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Upgrade to Storybook 9

- explicitly install @storybook/addon-docs
- Remove `docs: { autodocs: true }` in favor of `tags: ['autodocs']`
- `@storybook/manager-api` -> `storybook/manager-api`
- `@storybook/preview-api` -> `storybook/preview-api`
- `@storybook/theming/create` -> `storybook/theming/create`
- `@storybook/addon-viewport` -> `storybook/viewport`
- `@storybook/addon-actions` -> `storybook/actions`
  - Update `selectedPanel` from `actions` -> `storybook/actions/panel`
- `@storybook/blocks` -> `@storybook/addon-docs/blocks`
- Remove unused dependencies
- Upgrade vite
- patch @storybook/react (https://github.com/storybookjs/storybook/issues/31527)
- Remove dependence on `pharos-site` styles from the Storybook config

**Additional context**

- We will need to keep an eye on the patch file for `@storybook/react` (storybookjs/storybook#31527) and remove it when the issue is resolved to avoid maintenance burden later on

### DIFF
--- a/.storybook/styles.scss
+++ b/.storybook/styles.scss
@@ -1,0 +1,5 @@
+@use '@ithaka/pharos/lib/utils/scss/mixins';
+
+.hide-in-storybook {
+  @include mixins.hidden;
+}
--- a/packages/pharos-site/static/styles/global.scss
+++ b/packages/pharos-site/static/styles/global.scss
@@ -274,10 +274,6 @@ p.alert-example__content + p.alert-example__content {
   grid-column: span 4;
 }
 
-.hide-in-storybook {
-  @include mixins.hidden;
-}
-
 //  hide something from sighted users but still be accessible to screen readers
 .visually-hidden {
   @include mixins.hidden;
--- a/packages/pharos/src/components/tabs/pharos-tabs.wc.stories.jsx
+++ b/packages/pharos/src/components/tabs/pharos-tabs.wc.stories.jsx
@@ -1,4 +1,4 @@
-import { action } from '@storybook/addon-actions';
+import { action } from 'storybook/actions';
 import { html } from 'lit';
 
 import { configureDocsPage } from '@config/docsPageConfig';
--- a/packages/pharos/src/utils/scss/_variables.scss
+++ b/packages/pharos/src/utils/scss/_variables.scss
@@ -1,4 +1,4 @@
-@use '../../styles/variables';
+@use '../../styles/_variables.scss' as variables;
 
 $breakpoints: (
   'small': (
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -7,12 +7,9 @@ const config = {
   stories: [],
   addons: [
     getAbsolutePath('@storybook/addon-a11y'),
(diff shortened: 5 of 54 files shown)


packages/pharos/src/components/tabs/pharos-tabs.wc.stories.jsx
```diff
@@ -1,4 +1,4 @@
-import { action } from '@storybook/addon-actions';
+import { action } from 'storybook/actions';
 import { html } from 'lit';
 
 import { configureDocsPage } from '@config/docsPageConfig';
```
shoupeva-ithaka, line 4: I see `selectedPanel: 'addon-controls'` is left unchanged below, not sure if that needs to be updated similar to how 'addon-actions' was.

shoupeva-ithaka (approved):

.storybook/styles.scss
daneah, line 1: All of this transferred or duplicated from the `pharos-site`'s `global.scss`.

packages/pharos-site/static/styles/global.scss
daneah, line 1: Things removed here either weren't used at all or were only used within Storybook and moved to `.storybook/styles.scss`.

packages/pharos/src/utils/scss/_variables.scss
```diff
@@ -1,4 +1,4 @@
-@use '../../styles/variables';
+@use '../../styles/_variables.scss' as variables;
```
daneah, line 1: Under the newest version of Vite, naming conflicts started to occur because we have both a `_variables.scss` and a `variables.css` file in the `pharos/lib/styles/` directory, and Vite started resolving the CSS file instead of the SCSS file first. This explicit naming ensures the intended file is imported here.

packages/pharos/src/components/tabs/pharos-tabs.wc.stories.jsx
daneah: Ah thanks, yep needs to be updated. I'll double check for other instances and update that.

shoupeva-ithaka: Yeah, I saw some others too but all had the 'addon-controls' so if you grep that you should find them

sirrah-tam (approved):

packages/pharos/src/components/tabs/pharos-tabs.wc.stories.jsx
daneah: Eh, I think these need to stay as-is after looking further. They're not renamed, and they're still necessary AFAICT.

Issue #856, opened 2025-01-27
Sunset React wrappers


daneah:
**The problem**

React 19 [supports web components natively](https://custom-elements-everywhere.com/libraries/react/results/results.html). We should avoid expending further effort in supporting the React wrapper and the React Storybook, as it duplicates that effort and can create divergence.

**The solution**

- Remove React wrapper
- Remove React Storybook
- Remove React dependencies (unless somehow needed for Gatsby build)
- Streamline Storybook config to a single Storybook

This obviates the need for:

- #855 
- #854
- #812 
- #706

PR #991, opened 2025-06-20
Convert stories to TypeScript

### LINKED ISSUE #447: Convert component stories to TypeScript
**The problem**

Type safety is generally a productivity tool to aid in writing new code and avoiding regressions in existing code. Storybook is compatible with TypeScript and it would be nice to avoid type mismatches while authoring component stories.

**The solution**

Convert the current JSX story files to TSX and add type declarations accordingly.

**Additional information**

Patterns for using TypeScript for CSF aren't yet documented. [This StackOverflow issue](https://stackoverflow.com/questions/72549428/typescript-support-in-csf3) gives some leads on how to go about it.

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Resolves #447 

- Rename web component story files to `.ts`
- Rename React story files to `.tsx`
- Consistently use `Meta` and `StoryObj` to type stories
- Create `ComponentArgs` + `StoryArgs` convention in `storyArgs.ts` modules to specify component-level and story-level arguments (many of these are empty, but prefer the consistency)
- Update component creation templates in the `pharos-cli`
- Fix all type errors arising

**Additional context**

This required moving quite a few things from `.storybook/` into a new `src/utils/_storybook/` area because `.storybook/` is outside the build root for TypeScript. If there are better approaches to this, I'm very open to them (perhaps in a follow-on change). Probably could do it with an additional build config and some pathing magic, somehow, but wanted to do the bulk of the work first.

A nice benefit of this change + an upgrade to the linting dependencies is that we get quite a bit better linting on the web component stories.


packages/pharos/src/components/alert/storyArgs.ts
```diff
@@ -1,12 +1,23 @@
...
+  status: AlertStatus;
+  closable?: boolean;
+}
+
+export type StoryArgs = ComponentArgs & {
+  text: string;
+};
+
+export const defaultArgs: StoryArgs = {
+  status: 'base' as AlertStatus,
```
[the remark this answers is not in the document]
daneah: This is intentional at the moment; I'm however not positive whether there's value to having a `base` status in the future since it's not used in practice.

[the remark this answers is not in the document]
michael-iden: What is the intent behind the default status being an unsupported option? Like make sure all the Stories end up changing it to one of `('info', 'success', 'warning', 'error')`? Or do we want to see if an error is thrown?

https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/alert/pharos-alert.ts#L85

packages/pharos/src/components/alert/storyArgs.ts
```diff
@@ -1,12 +1,23 @@
+import type { AlertStatus } from '../../components/alert/pharos-alert';
+
 export const argTypes = {
   status: {
     options: ['info', 'success', 'warning', 'error'],
```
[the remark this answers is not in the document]
daneah: See other resolved thread about this.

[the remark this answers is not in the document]
michael-iden: I was kinda curious about this being intentional and asked on the other thread

> What is the intent behind the default status being an unsupported option? Like make sure all the Stories end up changing it to one of ('info', 'success', 'warning', 'error')? Or do we want to see if an error is thrown?
> 
> https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/alert/pharos-alert.ts#L85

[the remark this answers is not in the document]
daneah: Yeah, the intent was to have some kind of "placeholder" view so that instead of generating console errors only, there was some kind of visual indicator. This pattern didn't really take off in Pharos at large, and I think we want to oust it in the long term, but don't feel it's appropriate to manage here in this PR.

packages/pharos/src/components/alert/storyArgs.ts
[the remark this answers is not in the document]
daneah: Yeah, the intent was to have some kind of "placeholder" view so that instead of generating console errors only, there was some kind of visual indicator. This pattern didn't really take off in Pharos at large, and I think we want to oust it in the long term, but don't feel it's appropriate to manage here in this PR.

michael-iden (approved):

packages/pharos/src/components/button/storyArgs.ts
```diff
@@ -1,33 +1,64 @@
 import { iconNames } from '../../utils/iconNames';
+import type { PharosButtonProps } from '../../react-components/button/pharos-button';
+import type { LinkTarget } from './pharos-button';
+
+export interface ComponentArgs {
```
brentswisher, line 4: I wonder if we are already using the `PharosButtonProps` like this, could be just do something like:
`export interface ComponentArgs extends PharosButtonProps {}`

And use the whole thing, instead of manually recreating it?

brentswisher: 🤔 Alternatively, if we don't want to rely on the generated React types, we could do something like this, instead of creating a `ComponentArgs` interface:
`export interface StoryArgs extends Pick<PharosAlert, 'status' | 'closable'> {`

and then in the story it would change from `satisfies Meta<ComponentArgs>` to `satisfies Meta<PharosAlert>`

Using Pick instead of the React types means we still need to manually set the keys to pull from the component class, but we at least can still use the types, so that the story types and component types don't get out of sync

daneah: FWIW this is the only story where the component props are currently imported into `storyArgs`, because I needed the types to agree for the `icon*` props and wasn't readily finding another way to do so. I think what you're proposing sounds like a potentially good improvement, but wondering if it's maybe separable from this already large change?

daneah: In fact, I just pushed a change for these to use `IconName` instead—so there are no more uses of `<Component>Props` in `storyArgs` currently. I still like your idea and maybe we can create a separate issue for looking into it?

packages/pharos/src/components/image-card/PharosImageCard.react.stories.tsx
```diff
@@ -519,8 +523,8 @@ export const SelectableCollection = {
           title="Selected Disabled"
           link="#"
           variant="selectable-collection"
-          disabled="true"
-          selected="true"
+          disabled
+          _isSelected
```
brentswisher, line 527: This looks odd to me, in that we are setting what looks to be a private attribute via the public interface? Non-blocking just calling it out

brentswisher (approved):

packages/pharos/src/components/image-card/PharosImageCard.react.stories.tsx
daneah: Fair call—what I was running into here was a TS error in VSCode:

```
Type '{ children: Element; title: string; link: string; "source-type": string; variant: "selectable"; disabled: true; selected: true; }' is not assignable to type 'IntrinsicAttributes & PharosImageCardProps'.
  Property 'selected' does not exist on type 'IntrinsicAttributes & PharosImageCardProps'. ts(2322)
```

where `PharosImageCardProps` only has:

```typescript
export interface PharosImageCardProps extends DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement> {
    ...

    /**
     * Indicates if the image card is selected
     */
    _isSelected?: boolean;

    ...
}
```

This is due to the fact that we have `_isSelected` as a `public` property in the component source code. However, the actual TypeScript compilation and linting do work even if I keep `selected` instead of `_isSelected`.

I'd personally vote that we keep this as-is, and work toward making the public property name change to `selected` in a next major version (and optionally introducing a forward-compatible property in the meantime). I'll open an issue for this as well.

daneah: That said, we're likely to get rid of React altogether in the next major version, as React 19+ natively supports web components. (#856)

PR #1028, opened 2025-07-21
feat(button): add pressed/selected state styling

### LINKED ISSUE #995: Button: Add a selected/pressed state
**The problem**
Currently, we don't have a way of indicating to the user that a button is pressed, especially if we need to show a single button and not a toggle button group. This has multiple use cases including in toolbars and forms.

**The solution**
Add selected states for the secondary and subtle button variants. 

_Button Secondary - Selected_

Image
Image
Image

Border: 1 px
Color: $pharos-color-interactive-secondary
Transition: $pharos-transition-base
Border-Radius: 2px

_Button Subtle - Selected_

Image
Image
Image

Fill: $pharos-color-ui-20 ($pharos-color-marble-gray-94)
Icon/Text: $pharos-color-engineering-orange (#C61515)
Transition: $pharos-transition-base
Border-Radius: 2px

We also need to update the Secondary Button hover state to add a fill of marble-gray-97 to align between the standard button and the toggle button.

Image

**Alternatives considered**
We considered the toggle button group, but the visual constraints of the toggle group make it most useful for specific contexts (moving between views in a page) and not for turning functions on and off. That being said, we want to align with the toggle-button's aria-pressed state for colors.

**Additional information**

We also would like to add the following information to the Pharos site.

Selected State:

Use the selected state of a button to indicate to a user that a button is active, usually when making a selection within the experience. This should be used sparingly and only if clicking the button directly affects the experience within the page. 

Do not apply this state if the button redirects a user to a new page. 

If the button contains text, we recommend using the secondary button styling. If the button is icons-only, we recommend using the subtle button styling. Avoid applying this state with the primary button styling to avoid creating confusion with multiple primary CTAs. 

If there is a default selection upon page load, consider using the toggle button group instead. 

Examples include a button to apply italics in a text editor or choosing between two different options in a form with progressive disclosure.

kmundlamuri:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

Resolves #995 

**Additional context**
Add any other context here.

### DIFF
--- a/packages/pharos/src/components/button/pharos-button.scss
+++ b/packages/pharos/src/components/button/pharos-button.scss
@@ -159,7 +159,7 @@
   );
 }
 
-:host([is-on-background][variant='secondary']:not([disabled],[selected]))
+:host([is-on-background][variant='secondary']:not([disabled],[a11y-pressed="true"]))
   #button-element:hover,
 :host([is-on-background][variant='secondary']:not([disabled],[selected]))
   #button-element:active {
--- a/packages/pharos/src/components/button/pharos-button.test.ts
+++ b/packages/pharos/src/components/button/pharos-button.test.ts
@@ -254,8 +254,12 @@ describe('pharos-button', () => {
       component.a11yPressed = 'true';
       await component.updateComplete;
 
-      expect(component.hasAttribute('a11y-pressed')).to.be.true;
+      const button = component.renderRoot.querySelector('#button-element') as HTMLElement;
+      const computedStyle = getComputedStyle(button);
+      
+      expect(component.getAttribute('a11y-pressed')).to.equal('true');
       expect(component.getAttribute('variant')).to.equal('secondary');
+      expect(computedStyle.color).to.not.equal('');
     });
 
     it('applies pressed styling for subtle variant', async () => {
@@ -263,8 +267,12 @@ describe('pharos-button', () => {
       component.a11yPressed = 'true';
       await component.updateComplete;
 
-      expect(component.hasAttribute('a11y-pressed')).to.be.true;
+      const button = component.renderRoot.querySelector('#button-element') as HTMLElement;
+      const computedStyle = getComputedStyle(button);
+      
+      expect(component.getAttribute('a11y-pressed')).to.equal('true');
       expect(component.getAttribute('variant')).to.equal('subtle');
+      expect(computedStyle.backgroundColor).to.not.equal('');
     });
   });
 
--- a/packages/pharos/src/components/button/pharos-button.wc.stories.ts
+++ b/packages/pharos/src/components/button/pharos-button.wc.stories.ts
@@ -53,8 +53,9 @@ export const Base: Story = {
 
 export const Variants: Story = {
   render: () => html`
-    <div style="display: grid; grid-gap: 2rem; grid-template-columns: repeat(3, 200px);">
+    <div style="display: grid; grid-gap: 2rem; grid-template-columns: repeat(4, 200px);">
       <div style="padding: 1rem; display: grid; grid-gap: 1.5rem;">
+        <h4>Normal</h4>
         <storybook-pharos-button name="primary">Primary</storybook-pharos-button>
         <storybook-pharos-button name="secondary" variant="secondary"
           >Secondary</storybook-pharos-button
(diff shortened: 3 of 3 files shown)


packages/pharos/package.json
```diff
@@ -58,6 +58,7 @@
   "dependencies": {
     "@floating-ui/dom": "^1.7.2",
     "@ithaka/focus-trap": "2.0.1",
+    "@ithaka/pharos": "file:.yalc/@ithaka/pharos",
```
daneah:
```suggestion
```

packages/pharos/src/components/toggle-button-group/pharos-toggle-button.ts
```diff
@@ -20,7 +20,7 @@ export class PharosToggleButton extends PharosButton {
    * @attr selected
    */
   @property({ type: Boolean, reflect: true })
-  public selected = false;
+  public override selected = false;
```
daneah: Perhaps this no longer needs to be present, if inheriting it from the superclass now?

packages/pharos/package.json
kmundlamuri: Thanks Dane, i removed the line of code

packages/pharos/src/components/toggle-button-group/pharos-toggle-button.ts
kmundlamuri: Agreed! Removed the duplicate selected property - the toggle button now inherits it from PharosButton. This fixes the override conflict and eliminates code duplication.

brentswisher:
Preview available here, we can trigger this manually now but doesn't seem to add a status check, which makes sense: https://60919c26122bd50039b34644-uczdkxjbjc.chromatic.com/?path=/story/components-button--selected

kmundlamuri:
> Hey @kmundlamuri - sorry again for the delay, this sparked quite the discussion with the maintainers deciding the best way forward 😄
> 
> Here is what we are thinking:
> 
> * Instead of adding a new attribute for `selected` we would like to use the existing `a11yPressed` attribute, and add the styles based on this. (Since we would never want `selected` and `a11yPressed` to be out of synch, it makes more sense to lean into what we already have.)
> * This will be a little more complicated as the `toggle-button-group` has a `toggle-button` which extends from the base button, but needs to maintain separate styles.
> 
> So to move forward, we want to rework you PR to add styles to the existing `a11yPressed` attribute, and then we will likely also need to adjust the CSS of the toggle button to override those so that it remains styled as it currently is.
> 
> I know that is a little more work than you starting with, but I'm happy to pair with you Friday afternoon if it starts to get tricky, just let me know!

Thanks @brentswisher for the update and context! I’ve updated the PR to use the existing a11yPressed attribute for styling as discussed, and adjusted the toggle button CSS so that its current styling is preserved.

Please let me know if you’d like me to walk through the changes or if any further tweaks are needed.

sirrah-tam (approved):
Validated that the selected buttons convey the selected state programmatically (ie. passing down the `aria-pressed="true"` to the button) and that the visual styles of the toggle buttons are maintained while the new styles for selected buttons are added.

packages/pharos/src/components/button/pharos-button.scss
```diff
@@ -151,16 +151,25 @@
...
-:host([is-on-background][variant='secondary']:not([disabled])) #button-element:active {
+:host([is-on-background][variant='primary']:not([disabled])) #button-element:active {
   @include mixins.button-color(
     $color: var(--pharos-button-color-on-background-primary-text-hover),
     $background-color: var(--pharos-button-color-on-background-primary-background-hover),
     $border-color: var(--pharos-button-color-on-background-primary-border-hover)
   );
 }
 
+:host([is-on-background][variant='secondary']:not([disabled],[selected]))
```
brentswisher: I don't think we will want `selected` here?

packages/pharos/src/components/button/pharos-button.test.ts
```diff
@@ -242,6 +242,30 @@ describe('pharos-button', () => {
...
+      await component.updateComplete;
+      await expect(component).to.be.accessible();
+    });
+
+    it('applies pressed styling for secondary variant', async () => {
+      component.variant = 'secondary';
+      component.a11yPressed = 'true';
+      await component.updateComplete;
+
+      expect(component.hasAttribute('a11y-pressed')).to.be.true;
```
brentswisher: These bottom two tests don't seem to be testing much other than the attributes that are being set are in fact set, which don't add much value. I wonder if we can test the styles are set as expected or something like that?

packages/pharos/src/components/button/pharos-button.wc.stories.ts
```diff
@@ -289,3 +289,39 @@ export const Forms: Story = {
     </div>
   `,
 };
+
+export const Selected: Story = {
```
brentswisher: I'm not sure have a new story for "selected" makes sense here any more, maybe we could update the variants to have a toggle-able state that sets/unsets a11y-pressed>

kmundlamuri:
> Left a few comments on things

Thanks @brentswisher I have made recommended changes, please let me know you thoughts

brentswisher:
@joiechang @sirrah-tam 
I kind of anticipated when I press on a button, it would show the pressed styles for as long as I am holding my mouse on it, but maybe I am misunderstanding the intent? I know we also want people to make it so the press "sticks" by toggling the a11y-pressed attribute, but should it always be styled as pressed too?

PR #1004, opened 2025-06-30
fix(sheet): Adding code to stop click from propagating 

phosphotungstic:
**This change:** (check at least one)

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves maintainability
- [ ] Improves documentation
- [ ] Is a release activity

**Is this a breaking change?** (check one)

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Title of this pull request clear, concise, and indicative of the issue number it addresses, if any?
- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?
- [ ] Component status page up to date?

**What does this change address?**
Fixes #1002 

**How does this change work?**
Adds `preventDefault` and `stopPropagation` so when you click on the overlay, nothing behind it gets clicked.

PR #1030, opened 2025-07-21
fix(sheet): Moving stop propagation into correct place

phosphotungstic:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [x] Changeset added?

In #1004, a change was pushed to fix a bug in the `Sheet` component where the overlay was able to be clicked through. The stopping of the event propagation was in the wrong place which made events inside of the `Sheet` stop. 

Puts code into the check that checks if the interaction was on the shadowed part of the overlay.


.changeset/forty-nights-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ithaka/pharos': minor
```
brentswisher: I think this would probably be a `patch` release, since it's fixing a bug

brentswisher (approved):

Issue #1040, opened 2025-08-06
Sheet: Focus management fallback on open


sirrah-tam:
**Expected behavior**
When the sheet opens, there should be a consistent behavior where keyboard focus moves, either to the sheet outer container, the handle, or somewhere where on next tab stop focus should be within the sheet. 

**Actual behavior**
When there is no specified focus target, focus remains on the triggering element behind the opened sheet. Users are then required to navigate the rest of the page to get to the sheet. This is the current behavior that is seen within the Storybook example

**Steps to reproduce the issue**

1. Navigate to and activate the triggering element that opens the sheet component
2. Using the keyboard only, attempt to navigate to the sheet component
3. Observe where focus moves is located
4. Alternatively, once the sheet appears, use the browser console to inspect the current active element using `document.activeElement`

**Screenshots or code**

<img width="459" height="338" alt="screenshot of sheet component with focus indication still on triggering element" src="https://github.com/user-attachments/assets/30db6df3-6871-4f14-ad0f-534e3ffcbeaf" />

**Pharos version**
14.16.2

**Additional information**
Passing elements with the `data-sheet-focus` attribute into the sheet's slot does work as expected, but the documentation around use of this isn't clear. Additionally, if nothing is used there doesn't appear to be an applicable element in the `focusable` list that helps here.

PR #1041, opened 2025-08-06
Leverage shadowRoot to target the focuse element of the sheet handle

afiasco:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [ ] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

**Additional context**
Add any other context here.

### DIFF
--- a/packages/pharos/src/components/sheet/pharos-sheet.ts
+++ b/packages/pharos/src/components/sheet/pharos-sheet.ts
@@ -119,7 +119,6 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
         this._sheetContent.style.height = this.expanded
           ? this.MAX_EXPAND_PERCENTAGE
           : this.MIN_EXPAND_PERCENTAGE;
-        this._focusContents();
       } else {
         this._returnTriggerFocus();
       }
@@ -196,6 +195,7 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
         this._sheetContent.style.height = this.expanded
           ? this.MAX_EXPAND_PERCENTAGE
           : this.MIN_EXPAND_PERCENTAGE;
+        this._focusContents();
       }
     }
   }
@@ -312,7 +312,8 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
 
   private async _focusContents(): Promise<void> {
     this._currentTrigger = this.ownerDocument.activeElement;
-    const focusElement = this.querySelector(FOCUS_ELEMENT);
+    const focusElement = this.shadowRoot?.querySelector(FOCUS_ELEMENT);
+
     if (focusElement) {
       await 0;
       (focusElement as HTMLElement).focus();
@@ -398,7 +399,7 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
                 @mousedown=${this._handleMouseDragStart}
                 @mouseup=${this._handleDragEnd}
               >
-                <div class="sheet__handle"></div>
+                <div class="sheet__handle" tabindex="-1" data-sheet-focus></div>
               </div>
               ${this._renderCloseButton()}
               <div class="sheet__body">


sirrah-tam (approved):

packages/pharos/src/components/sheet/pharos-sheet.test.ts
```diff
@@ -83,7 +83,7 @@ describe('pharos-sheet', () => {
     expect(mockHandler.called).to.be.false;
   });
 
-  it('returns focus back to the trigger element when close', async () => {
+  it('focus moves to the sheet after opening and return back to the trigger element when closed', async () => {
```
brentswisher:
```suggestion
  it('focus moves to the sheet after opening and returns back to the trigger element when closed', async () => {
```

packages/pharos/src/utils/focusable.ts
```diff
@@ -26,7 +26,8 @@ const focusable = `
   [data-pharos-component="PharosDropdownMenuNavCategory"],
   [data-pharos-component="PharosSidenavLink"][href],
   [data-pharos-component="PharosToast"],
-  [data-pharos-component="PharosBreadcrumbItem"][href]
+  [data-pharos-component="PharosBreadcrumbItem"][href],
+  [data-sheet-handle]
```
brentswisher: Can we limit this to the sheet component, like the ones above this?

afiasco: giving this more specificity seems to break the focus changes 🫤 trying to figure out whats going on there - do you have any thoughts?

afiasco: @brentswisher after chatting with @phosphotungstic I wonder if we should remove this from `focusable` because I think we're using it a bit differently than the others on this list?

and instead just add an additional `const FOCUS_HANDLE = [data-sheet-handle]` into the component that we'll pass [here](https://github.com/ithaka/pharos/blob/a84e29fb0e5404f6a7d1409819ec3547fcaf9792/packages/pharos/src/components/sheet/pharos-sheet.ts#L320)

brentswisher: 😄 Not really opposed, but could you explain a little more how it's being used differently from the others, just so I make sure I understand what its doing?

phosphotungstic: Yeah so when we looked into it, it looked like `focusable` was mainly being used inside of Pharos components to find _other_ Pharos components to scope to. So stuff like `PharosBreadCrumbItem` wasn't the actual component that it was querying _inside of_ vs querying _for_.
So when we tried to add `[data-pharos-component="PharosSheet"]` it didn't work because:
- When it tries to use that query selector, when we tried to use it with `this.shadowRoot`, it doesn't exist since the `shadowRoot` is _inside_.
- When we tried doing it in the parent component and such, it also doesn't work because we can't actually query for `[data-sheet-handle]` (or anything for the matter) inside of shadow roots.

afiasco: this change has been pushed up - if it looks alright!

brentswisher: Not adding the `[data-pharos-component="PharosSheet"]` makes sense to me now, thank you!

brentswisher: I do wonder if the change to focus on the handle specifically instead of adding it to the list of possible focusable elements is desired though? 

Does that would mean it always focus on the handle when it wasn't given a focus element, even if there are other focusable elements inside it could default to? If so, is that the desired behavior @sirrah-tam?

phosphotungstic: > Does that would mean it always focus on the handle when it wasn't given a focus element, even if there are other focusable elements inside it could default to? If so, is that the desired behavior

I believe this is true but I think that's fine? Since we always know that the handle will be there and if the user wants to choose a different focusable element, they can just use `data-sheet-focus`.

afiasco: iirc thats the intended behavior - if the consumer doesnt' provide a focusable element it should default goto the handle

sirrah-tam: Yep, this is what we're aiming for to help resolve the issue #1040 which calls out that we don't have a fallback focus target when the consumer doesn't include something with `data-sheet-focus` or something that is considered `focusable` via that library.

brentswisher: >or something that is considered focusable via that library.

👍 @sirrah-tam Just want to be explicit, with this change even if there was something it considered focusable, it will always focus on the handle element instead

That's the part I wanted to verify

sirrah-tam: Ah! You're right! Missed that part of the change, we're getting rid of the `focusable` conditional in this case. I'm still less familiar with the purpose of that `focusable` library and it may just be an artifact of the Pharos Modal which the Sheet was built off of. I'd say as long as folks are able to overwrite the focus target with the `data-sheet-focus` we should be okay.

afiasco: @sirrah-tam @brentswisher does this indicate we're comfortable leaving the `FOCUS_HANDLE` constant in the modal file? Or do we prefer it gets moved back to the `focusable` list? It's still not entirely clear to me the way we want to move forward.

sirrah-tam: I'm comfortable leaving the `FOCUS_HANDLE` constant and moving away from using the focusable list!

PR #1051, opened 2025-08-19
Sheet: Update mobile sticky sheet

vedanthkallakuri:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?

This change adds a sheet type called OmitOverlay, which allows the sheet to start at a custom docked height and expand to an expanded height on drag or button click. It allows for the start and expanded pixel heights to be specified in the component itself to make the sheet more customizable. The new RT mobile sticky sheet button ([figma](https://www.figma.com/design/wQglDD1DQuVhLfzodPhjE8/Research-tool---Mobile?node-id=6279-45813&p=f&t=Jc1N0orfseGHUqFp-0)) intends to use the sheet in this new way. Thus, we are aiming to add this new sheet feature to allow for that functionality.


daneah:
@vedanthkallakuri could you please update the pull request description with details about this change? Thank you!

packages/pharos/src/components/sheet/pharos-sheet.wc.stories.ts
daneah, line 1: For now you will also need to add a React story as well.

packages/pharos/src/components/sheet/pharos-sheet.wc.stories.ts
```diff
@@ -85,3 +85,18 @@ export const LongContent: Story = {
     </div>
   `,
 };
+
+export const omitOverlay: Story = {
```
daneah:
```suggestion
export const OmitOverlay: Story = {
```

packages/pharos/src/components/sheet/pharos-sheet.wc.stories.ts
```diff
@@ -85,3 +85,18 @@ export const LongContent: Story = {
...
+  render: () => html`
+    <div>
+
```
daneah:
```suggestion
```

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -375,41 +448,51 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
...
-        </div>
+          </div>
+        </focus-trap>
       </div>
     `;
   }
-}
+
+  protected override render(): TemplateResult {
+    // omitOverlay -> do not render the overlay wrapper
```
daneah:
```suggestion
```

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -375,41 +448,51 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
       : nothing;
   }
 
-  protected override render(): TemplateResult {
+  protected renderSheet(): TemplateResult {
+    const sheetDialog = this.omitOverlay ? 'sheet__dialog_no_overlay' : 'sheet__dialog';
```
daneah: From a BEM perspective, "no overlay" is more like a modifier, so I would update the name accordingly:

```ts
    const sheetDialog = this.omitOverlay ? 'sheet__dialog--no-overlay' : 'sheet__dialog';
```

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -159,6 +191,29 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
...
   }
 
+  private _normalizePx(value?: string, fallback = '60px'): string {
+    if (!value) return fallback;
+    const v = value.trim();
+    if (/^\d+px$/.test(v)) return v;
+    const n = parseInt(v, 10);
+    if (!Number.isNaN(n)) return `${n}px`;
+    return fallback;
+  }
```
daneah: I'm not sure I see the value of this—can you elaborate on its use?

vedanthkallakuri: I was using it to parse different types of pixel inputs, but now I realize that it is probably unnecessary. Just removed it to clean things up.

sirrah-tam (commented):
There doesn't appear to be a way to adjust the height of the sheet with only the keyboard. It looks like we're giving the handle some mouse action events, but it doesn't translate to what happens if a keyboard only user is interacting with it. We may need to turn that handle into an actual button for this functionality (or give it `role="button"` & `tabindex="0"` along with an accessible name). Visually, I think it can stay the same, but will need to have some of this interaction not be only done via mouse/touch actions.

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -73,6 +73,27 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
   @property({ type: String, reflect: true })
   public header = 'Sheet header';
 
+  /**
+   * Start height for omitOverlay mode, only accepts px (e.g. "200px")
+   * @attr start-height
+   */
+  @property({ type: String, attribute: 'start-height', reflect: true })
+  public startHeight = '60px';
```
ymouzakis: How do we feel about using pharos-spacing-tokens to define these pixel values instead?

I was thinking about users of the sheet being able to pass a number divisible by 0.25 and then that getting converted over to pharos-spacing-tokens.

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -24,6 +24,7 @@ const FOCUS_HANDLE = `[data-sheet-handle]`;
  *
  * @fires pharos-sheet-open - Fires when the sheet is about to open - cancelable
  * @fires pharos-sheet-opened - Fires when the sheet has opened
+ * @fires pharos-sheet-expanded- Fires when the sheet has expanded
```
daneah:
```suggestion
 * @fires pharos-sheet-expanded - Fires when the sheet has expanded
```

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -73,12 +74,33 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
   @property({ type: String, reflect: true })
   public header = 'Sheet header';
 
+  /**
+   * Indicates the minimim height for the sheet in pixels
```
daneah:
```suggestion
   * Indicates the minimum height for the sheet in pixels
```

packages/pharos/src/components/sheet/pharos-sheet.ts
```diff
@@ -278,26 +335,27 @@ export class PharosSheet extends ScopedRegistryMixin(PharosElement) {
...
         this.dispatchEvent(new CustomEvent('pharos-sheet-expanded', details));
         this.expanded = true;
       } else {
         if (this.expanded) {
-          this._sheetContent.style.height = this.MIN_EXPAND_PERCENTAGE;
+          this._sheetContent.style.height = this._getMinHeightStr();
           this.dispatchEvent(new CustomEvent('pharos-sheet-collapsed', details));
           this.expanded = false;
-        } else {
+        } else if(!this.docked) {
```
daneah: I suspect you way need to install the pre-commit hooks—I feel Prettier would reformat this code!

.changeset/fuzzy-owls-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ithaka/pharos': minor
+---
+
+Created new docked sheet and the option to removed the overlay
```
daneah:
```suggestion
Sheet now supports being `docked` or to `omitOverlay`
```

daneah (approved):
A suggestion for a bit more consumer-facing changeset entry, to take or leave. Looking good otherwise 😎

mtorres3 (approved):

PR #1066, opened 2025-09-09
Add Pill Component

### LINKED ISSUE #1056: Pill component
A pill component would allow multiple teams to adopt consistent behavior and stying for pills. Pills are a very common UI component across other design systems. Pills allow for information to be displayed and/or interacted with in controlled ways. 

**Landscape analysis**

Here are some DS documentation that can be reviewed.
• [Horizon Design System](https://horizon.servicenow.com/workspace/components/now-pill)
• [Workday Design System](https://canvas.workday.com/components/indicators/pill/)

**Additional considerations**

[Here is Pharos documentation](https://docs.google.com/document/d/1FXVY2qbMnN4Ofrf69ZzWSOqGNfzWhhNvtABbcod1Y7o/edit?tab=t.0#heading=h.68w7cixw5hi7). 

[Here is the Pharos figma](https://www.figma.com/design/sPvnmrxhsJpQqMYrQKwXve/Pharos-UI-Kit?node-id=23492-17436&t=VaWsWgAA0jcVGvOf-4)

brentswisher:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Closes #1056. 

Adds a new Pill component to Pharos. Pills can be informational or dismissible, if they are dismissible they render as a button with a close icon on the right otherwise they render as a div.

**Additional context**
- The text content will be truncated with ellipses if it is too long, per design pills should never line wrap
- I had to do some dancing with the storyArgs to get all the controls working properly in both React and WebComponents, which is why you will see both a `iconLeft` and a `icon-left` - this looks a little messy in the code but should be clearer for consumers
- There is a stacked PR (#1084) for adding documentation to pharos.jstor.org that should be merged before this is released.


.changeset/ninety-moments-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ithaka/pharos': minor
+---
+
+Add new Pill component
```
daneah:
```suggestion
Add new Pill component for displaying discrete selections
```

daneah (commented):
It appears that pills with icons are misaligned to the other pill types:

Image

packages/pharos/src/components/pill/pharos-pill.wc.stories.ts
```diff
@@ -0,0 +1,331 @@
...
+        <br />
+      </div>
+
+      <div
+        style="width:300px; padding:2rem; border:1px solid var(--pharos-color-black); margin-top:2rem; resize:horizontal; overflow:auto;"
+      >
+        <storybook-pharos-heading level="2" preset="5"
+          >Truncating Pills (Container)</storybook-pharos-heading
+        >
+        <p>Pill will also be truncated when the are longer than their container.</p>
```
daneah:
```suggestion
        <p>Pills are truncated when they are longer than their container.</p>
```

packages/pharos/src/components/pill/PharosPill.react.stories.tsx
```diff
@@ -0,0 +1,471 @@
...
+          border: '1px solid var(--pharos-color-black)',
+          marginTop: '2rem',
+          resize: 'horizontal',
+          overflow: 'auto',
+        }}
+      >
+        <PharosHeading level={2} preset="5">
+          Truncating Pills (Container)
+        </PharosHeading>
+        <p>Pill will also be truncated when the are longer than their container.</p>
```
daneah:
```suggestion
        <p>Pills are truncated when they are longer than their container.</p>
```

sirrah-tam (approved):

packages/pharos/src/components/pill/pharos-pill.ts
```diff
@@ -0,0 +1,166 @@
...
+        console.log(e);
+        throw new Error(`Could not get icon named "${this.iconLeft}"`);
+      }
+    }
+  }
+
+  // We need the normal icon proportions, but at smaller sizes, so we have to manually render the svg instead of using pharos-icon
+  private _renderIcon(icon: string, title: string, size: string): TemplateResult | undefined {
+    const iconSize = size === 'small' ? 12 : 16;
+    const iconBlob = atob(icon);
```
lizlove, line 101: Curious why you needed an atob function + unsafeSVG?

lizlove: Not a blocker at all just wanted to learn more.

brentswisher: Yeah, it's a bit different for sure, and in fairness, I stole it a bit from the PharosIcon component, but I'll explain it as best I can 😄 . The icon passed into the render is from an `import` statement on line 87 above, this becomes a base64 encoded file at some point when it's built - either Vite or rollup I think.

Lit's [unsafeSVG](https://lit.dev/docs/templates/directives/#unsafesvg) directive is similar to something like Reacts `dangerouslySetInnerHTML` - it just output the string you are passing out without escaping anything, but it needs a plain string input.

So `atob` converts the imported file from base64 to just a plan text string that contains the code for the SVG. Then `unsafeSVG` puts the SVG into the DOM. Since we control the source of the SVG (its the icon file), it's not particularly unsafe to use.

All of that was only necessary (Instead of just using a Pharos icon), because the Pharos icon hardcodes the height and width to 24 and we needed it to be 12/16 here. But to maintain the proportions of the icon, we need to keep the the viewbox at 24x24.

PR #979, opened 2025-06-10
ImageCard: refactor to remove duplicate link & address accessibility issues

sirrah-tam:
**This change:** (check at least one)

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves maintainability
- [ ] Improves documentation
- [ ] Is a release activity

**Is this a breaking change?** (check one)

- [ ] Yes
- [x] No
- [x] Maybe?

**Is the:** (complete all)

- [x] Title of this pull request clear, concise, and indicative of the issue number it addresses, if any?
- [x] Test suite(s) passing?
- [ ] Code coverage maximal?
- [ ] Changeset added?
- [ ] Component status page up to date?

**What does this change address?**
This change addresses #353 by removing the link that wraps the image within the image card. Also removes the keyboard navigation logic and leaves the checkbox in the DOM and only visually hides/shows it when in focus/hover. Also addresses accessibility issues with storybook examples.

Resolves #353 
Resolves #972

**How does this change work?**
This change removed the pharos-link from wrapping the image and created a click handler on the image to navigate to item's link. Added styles to convey interactivity.

PR #1069, opened 2025-09-12
Image card: Subtle-select doesnt navigate when checkbox is checked

### LINKED ISSUE #1065: Image Card: Clicking subtle-select navigates when checkbox is selected
**Expected behavior**
Describe what you would expect to happen.
When clicking on the image card while the checkbox is selected on the subtle-select variant, the checkbox should be deselected and the selected state removed. It should NOT navigate when clicked.

**Actual behavior**
When clicking on the image card while the checkbox is selected on the subtle-select variant the user is navigated to the linked page. 

1. Select the checkbox on an image-card subtle-select variant
2. Click on the card
3. Observe the page navigates to the linked content

**Pharos version**
14.15.0 +

**Your environment**

- OS: macOS
- Browser: Chrome 
- Version: 139.0.7258.157

**Additional information**
Fixing this issue should not break the other selectable type image card states

afiasco:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

resolves #1065 

When using the subtle-select image card state, if the checkbox was selected it would navigate when clicking on the card. The expected behavior is that clicking on the card when the checkbox is selected it would de-select the checkbox and NOT navigate. There are also some refactoring changes to make the function more readable and to remove a previously added nested click-handler. Thus this change:

- removes the `_handleImageClick` function and it's use as a nested handler from the component
- Adds the logic from `_clickImage` into the `_cardToggleSelect` click handler, which manually sets the url to the document object when clicking on the card
  - added conditional statement to check if the checkbox is checked to preventDefault and propagation.
  - added additional logic which prevents this from occurring on a `pharos-link`
- Refactors some variables and logic to be clearer

When clicking a selected image card it was navigating the user to the provided link instead of de-selecting the checkbox and changing the selected state. 
- Get the current state of the checkbox
- Added it to previous logic, where if `true` it does not navigate
- Added additional logic to check if the `event.currentTarget` is a `pharos-link`, if so it does not manually set the `document.location.href` 

**Additional context**
This was a breaking change introduced in pharos release `4.15.0` and this [pull-request](https://github.com/ithaka/pharos/pull/979)


packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -514,6 +515,8 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
           },
         })
       );
+    } else if (!isTitleClick) {
+      document.location = this.link;
```
sirrah-tam:
```suggestion
      document.location.href = this.link;
```
Reading up on this to make sure its the most appropriate way to navigate to pages using Javascript, this makes it more explicit about the purpose and ensures support across all browsers.

packages/pharos/src/components/image-card/pharos-image-card.ts
```diff
@@ -495,13 +489,20 @@ export class PharosImageCard extends ScopedRegistryMixin(FocusMixin(PharosElemen
...
-    const checkboxClicked = event.currentTarget === this._checkbox;
+    if (this.disabled) {
+      return;
+    }
 
-    if (!this.disabled && (cardClicked || checkboxClicked)) {
-      // this is required to prevent navigation on the link click
+    const isCardClick = this._isSelectableViaCard() && event.target !== this._checkbox;
+    const isCheckboxClick = event.currentTarget === this._checkbox;
+    const isCheckboxChecked = this._checkbox?.checked ?? false;
```
phosphotungstic, line 498: I think it's more common to do something like `!!this._checkbox?.checked`

PR #1087, opened 2025-10-02
Sheet: docked overlay interaction

ymouzakis:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

This is fixing a bug in the sheet. The docked sheet should never fully close. Clicking on the overlay should collapse the sheet, not close it.  

This change also includes a refactor of the sheet when the overlay is omitted so that the overlay is still on the page, but events pass though. This was done so that making the transition between omitOverlay being true/false to be a smooth one and not jaring. 

In the overlay interaction handler check to see if the sheet is docked and collapse it or close it as necessary.


.changeset/warm-fans-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ithaka/pharos': minor
```
michael-iden: nit - Is this a bugfix more than minor improvement?

ymouzakis: I wouldn't think so? This is not a breaking change so it's not major. Is there something more appropriate? Patch?

daneah: Bug fixes should be patches, yes, unless they need to break the API to accomplish the fix in which case they should be major version bumps.

michael-iden (approved):

packages/pharos/src/components/sheet/pharos-sheet.scss
```diff
@@ -9,6 +9,11 @@
   contain: layout size style;
 }
 
+// :host([open]),
+// :host([docked]) {
+//   pointer-events: auto;
+// }
```
daneah:
```suggestion
```

michael-iden (approved):

PR #1152, opened 2025-12-16
Pagination: Add input variant

jialin-he:

- [x] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [ ] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Add a new variant called "input" for pagination. This allows the consumer to input the page number to jump to the page.

image

image

If it is the input variant, show an input box as the current page and emit an event when enter key down.

### DIFF
--- a/packages/pharos/src/components/pagination/pharos-pagination.scss
+++ b/packages/pharos/src/components/pagination/pharos-pagination.scss
@@ -30,13 +30,16 @@
   border: 1px solid var(--pharos-form-element-color-border-base);
   border-radius: var(--pharos-radius-base-standard);
   font: inherit;
+  font-variant-numeric: tabular-nums;
   color: var(--pharos-form-element-color-text-base);
   background: none;
   width: auto;
   box-sizing: border-box;
+  text-align: center;
 
   &:focus {
-    @include mixins.input-focus;
+    outline: 1px solid var(--pharos-color-focus);
+    border-color: var(--pharos-form-element-color-border-focus);
   }
 
   &:disabled {


packages/pharos/src/components/pagination/pharos-pagination.scss
```diff
@@ -8,12 +8,27 @@
...
   padding: 0 var(--pharos-spacing-1-x);
 }
 
+.pagination__input-wrapper {
+  display: inline-flex;
+  align-items: center;
+}
+
+.pagination__input {
+  height: 34px;
```
brentswisher: We usually try to avoid setting exact pixel sizes, I'm curious is this really _needs_ to be shorter/taller than our normal input?

packages/pharos/src/components/pagination/pharos-pagination.test.ts
```diff
@@ -152,4 +153,130 @@ describe('pharos-pagination', () => {
...
+        page-size="25"
+        variant="input"
+      ></test-pharos-pagination>
+    `);
+
+    const pageInput = component.renderRoot.querySelector('.pagination__input');
+    expect(pageInput).to.exist;
+  });
+
+  it('fires page-input event with clamped value on change', async () => {
```
brentswisher: I'm not sure I understand the "clamped value" ?

packages/pharos/src/components/pagination/pharos-pagination.test.ts
brentswisher, line 1: Can we add an accessibility test of the input variant too?

packages/pharos/src/components/pagination/pharos-pagination.ts
```diff
@@ -78,6 +96,15 @@ export class PharosPagination extends ScopedRegistryMixin(PharosElement) {
         `currentPage value '${this.currentPage}' is invalid. Can only be a number greater than or equal to 1`
       );
     }
+    if (changedProperties.has('currentPage')) {
+      this._pageInputValue = '';
+    }
+
+    if (changedProperties.has('variant') && this.variant && !VARIANTS.includes(this.variant)) {
+      throw new Error(
+        `${this.variant} is not a valid variant. Valid variants are: ${VARIANTS.join(', ')}`
```
brentswisher:
```suggestion
        `${this.variant} is not a valid Pharos pagination variant. Valid variants are: ${VARIANTS.join(', ')}`
```
Just to make the source more obvious from the console

packages/pharos/src/components/pagination/pharos-pagination.ts
```diff
@@ -143,11 +170,76 @@ export class PharosPagination extends ScopedRegistryMixin(PharosElement) {
...
+  }
+
   protected override render(): TemplateResult {
     return html`
       <div class="pagination__wrapper" role="navigation" aria-label="pagination">
         ${this._renderPrevLink()}
-        <span class="pagination__info">${this.currentPage} of ${this.totalPages}</span>
+        ${this.variant === 'input'
+          ? this._renderPageInput()
+          : html`<span class="pagination__info">${this.currentPage} of ${this.totalPages}</span>`}
```
brentswisher, line 286: Since we use this exact text inside `_renderPageInput` too, it might be worth extracting it to something like `_renderPaginationInfo` and using it in both places?

packages/pharos/src/components/pagination/pharos-pagination.ts
```diff
@@ -143,11 +170,76 @@ export class PharosPagination extends ScopedRegistryMixin(PharosElement) {
...
+
+    target.value = desiredPage.toString();
+    this._pageInputValue = target.value;
+
+    if (desiredPage === this.currentPage) {
+      return;
+    }
+
+    this.dispatchEvent(
+      new CustomEvent('page-input', {
```
brentswisher, line 235: I see this matches the existing events in the component, but I know for other custom events in Pharos we generally try to prefix them.  So this should maybe be `pharos-pagination-page-input`?  But then it wouldn't match the other pagination events, so maybe it it better to keep it like this 🤔   Wondering if @daneah has any thoughts on that?

packages/pharos/src/components/pagination/pharos-pagination.test.ts
jialin-he: actually it is testing if we have 10 pages in total and you enter 20 as the input, we should emit the event to jump to page 10 instead of 20. I can change the wording though

packages/pharos/src/components/pagination/pharos-pagination.scss
jialin-he: we actually don't need this. Removed it

packages/pharos/src/components/pagination/pharos-pagination.ts
jialin-he: we can only extract half of it otherwise we end up with something like this
image

jialin-he: If it is only half of it I feel it is not very useful to extract it

brentswisher: Yeah, good call, I misread that

packages/pharos/src/components/pagination/pharos-pagination.test.ts
jialin-he: done!

packages/pharos/src/components/text-input/pharos-text-input.ts
```diff
@@ -259,6 +259,7 @@ export class PharosTextInput extends ScopedRegistryMixin(FormMixin(FormElement))
           name=${this.name}
           .type=${this.type}
           .value=${this.value}
+          part="input"
```
brentswisher: This would open the `input` element inside `pharos-input` component's shadow dom up to _any_ CSS changes from outside, and isn't something we probably want to do. It probably makes sense to use a HTML input in the pagination control instead?

jialin-he: good idea! Done.

packages/pharos/src/components/pagination/pharos-pagination.ts
brentswisher: After sleeping on this, I do think we should keep it like you have it. If we want to make use of prefixed events, we can update all of them at once in a major release. I think not having consistency in the name of the events coming from the same component would be a really bad developer experience.

packages/pharos/src/components/pagination/pharos-pagination.ts
```diff
@@ -143,12 +196,98 @@ export class PharosPagination extends ScopedRegistryMixin(PharosElement) {
...
 
+  private _renderLastLink(): TemplateResult | typeof nothing {
+    if (this.variant === 'input' && this.currentPage < this.totalPages) {
+      return html`
+        <pharos-link
+          class="pagination__link last"
+          subtle
+          flex
+          href=""
+          aria-label="Go to last page"
```
brentswisher: I _think_ because we have the `pagination__visually-hidden` span we don't need the aria-label here

packages/pharos/src/components/pagination/pharos-pagination.ts
```diff
@@ -1,22 +1,31 @@
...
 /**
  * Pharos pagination component.
  *
  * @tag pharos-pagination
  *
+ * @fires first-page - Fires when the first page link is clicked
  * @fires prev-page - Fires when the previous page link is clicked
  * @fires next-page - Fires when the next page link is clicked
+ * @fires last-page - Fires when the last page link is clicked
+  * @fires page-input - Fires when a page number is submitted in the input variant
```
brentswisher:
```suggestion
 * @fires page-input - Fires when a page number is submitted in the input variant
```

packages/pharos/src/components/pagination/pharos-pagination.ts
```diff
@@ -105,6 +139,25 @@ export class PharosPagination extends ScopedRegistryMixin(PharosElement) {
...
   }
 
+  private _renderFirstLink(): TemplateResult | typeof nothing {
+    if (this.variant === 'input' && this.currentPage > 1) {
+      return html`
+        <pharos-link
+          class="pagination__link first"
+          subtle
+          flex
+          href=""
```
brentswisher, line 149: I think these blank hrefs might cause some accessibility issues - @sirrah-tam ?

I do see that the existing pagination also has them, so if it is an issue but we want to handle that in a followup I think that is fair.

packages/pharos/src/components/pagination/pharos-pagination.ts
jialin-he: removed

packages/pharos/src/components/pagination/pharos-pagination.ts
sirrah-tam: Ah, yeah, we typically want to have a null value in there, either a `#` or `javascript:;` kinda thing. But in this case it doesn't seem like it's causing any issues. Usually if there isn't an href that means its not in keyboard focus order, but here it appears to work correctly!

PR #1198, opened 2026-02-20
Add new focus trap controller

### LINKED ISSUE #777: Update to actively supported focus-trap library
**The problem**
We are using [@ithaka/focus-trap](https://github.com/ithaka/focus-trap) which is an out of date fork of a no longer maintained package to provide focus trapping for accessibility purposes. This package calls `document` directly in it's initialization which conflicts with the Gatsby SSR of pharos-site, requiring a clunky workaround on all pages for components which use it.

**The solution**
Update to a maintained library or write our own focus-trap inside of Pharos which is compatible with Gatsby SSR.

**Additional information**
- https://github.com/focus-trap/focus-trap looks to be well maintained
- #755 and #347 may impact this if the focus trap is no longer needed

brentswisher:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Closes #777 

This replaces the previous focus-trap implementation, with a new focus trap controller that uses the more actively maintained https://github.com/focus-trap/focus-trap library. It allows us to remove the previous implementation that relied on a forked package we are struggling to keep maintained.

**Additional context**

### DIFF
--- a/packages/pharos/src/components/popover/pharos-popover.ts
+++ b/packages/pharos/src/components/popover/pharos-popover.ts
@@ -26,7 +26,7 @@ const FOCUS_ELEMENT = `[data-popover-focus]`;
  *
  */
 export class PharosPopover extends ScopedRegistryMixin(FocusMixin(OverlayElement)) {
-  private _focusTrapController = new FocusTrapController(this, '.popover');
+  private _focusTrapController = new FocusTrapController(this, '.focus-trap');
 
   /**
    * Indicates the menu item is displayed on a dark background.
@@ -377,7 +377,7 @@ export class PharosPopover extends ScopedRegistryMixin(FocusMixin(OverlayElement
   protected override render(): TemplateResult {
     return html`
       <div
-        class="popover"
+        class="popover focus-trap"
         role="dialog"
         aria-label=${ifDefined(this.a11yLabel)}
         aria-labelledby=${ifDefined(this.labelledBy)}


packages/pharos/src/components/dropdown-menu/pharos-dropdown-menu.ts
```diff
@@ -34,9 +34,10 @@ const _allMenuItemsSelector = '[data-pharos-component="PharosDropdownMenuItem"]'
  *
  */
 export class PharosDropdownMenu extends ScopedRegistryMixin(FocusMixin(OverlayElement)) {
-  static elementDefinitions = {
-    'focus-trap': FocusTrap,
-  };
+  private _focusTrapController = new FocusTrapController(
+    this,
+    () => this.shadowRoot?.querySelector('.dropdown-menu__list') as HTMLElement | null
```
daneah: Can these all be replaced with assigned `@query` objects?

brentswisher: Yeah, that't probably true, I can take a look at that. I also have gone back and forth if it made more sense to use a "standard" class name like `focus-trap` for the element we want the focus trap on, instead of targeting specific classes in specific components.

brentswisher: So, I looked into this, and there is a timing issue. The `@query` objects don't exist until after first render, so we can't pass them in when the `FocusTrapController` is initialized.

Since the `FocusTrapController` is already getting an instance of the element as the first argument, we could assume the class name in the controller, and move the selection there. However, if we ever needed a component with 2 separate areas that had focus traps, we wouldn't have a way to do that. It also makes the coupling (which still exists) a little less explicit.

I think I will update it to accept a selector as the second argument. That feels like a good middle ground. It is a little easier in the components then passing in a `getContainer` function, but keeps the interface flexible for needing to select more than `.focus-trap`

Open to thoughts though!

daneah: That sounds good, thanks for the leg work on it!

packages/pharos/src/components/modal/pharos-modal.ts
```diff
@@ -38,9 +38,10 @@ export class PharosModal extends ScopedRegistryMixin(PharosElement) {
   static elementDefinitions = {
     'pharos-button': PharosButton,
     'pharos-heading': PharosHeading,
-    'focus-trap': FocusTrap,
   };
 
+  private _focusTrapController = new FocusTrapController(this, '.focus-trap');
```
daneah, line 43: I noticed in some places you add a `.focus-trap` class to an element and use that selector, while in others you use a selector for an existing class. Any methodology behind that that I'm unaware of?

brentswisher: it was unintentional, I updated it to use `.focus-trap`, but it looks like I missed the popover component.  I can update that, it doesn't _actually_ matter, but I thought keeping a consistent class name would make things easier to maintain.

brentswisher: [eca9a2d](https://github.com/ithaka/pharos/pull/1198/commits/eca9a2d4cea46c5c405c05c15d892bf951aa0aa2)

PR #1287, opened 2026-05-29
Add full width option to radio and checkbox components

### LINKED ISSUE #1281: Allow mechanism for <fieldset> within pharos-checkbox-group to expand
**The problem**
A full-width label for a checkbox in a checkbox group is currently not supported by pharos right now. This is because <fieldset> has [min-inline-size: min-content](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/min-inline-size) by default and we have no way to override it from the outside.

This prevents designs like this from happening

Image

brentswisher:

- [X] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Closes #1281 

This adds a new `full-width` attribute to the radio and checkbox components, which causes their labels to expand the full width of their container when provided. This allows adding additional styling to the checkbox host elements themselves, to denote selected state with additional visual changes.

**Additional context**
One side effect of this change I want to call out is it changes the default for the checkbox group and radio group inner containers to have a full width of 100%. There's a slight chance this could cause downstream issues, although since they don't provide any styling, I think it should be okay, but I wanted to call it out in case anyone has concerns. We could add "full-width" attributes to those components as well if there are concerns, but I wondered if that might be overkill

**Sceenshot**
Screenshot 2026-05-29 at 3 48 11 PM

### DIFF
--- a/packages/pharos/src/components/checkbox/PharosCheckbox.react.stories.tsx
+++ b/packages/pharos/src/components/checkbox/PharosCheckbox.react.stories.tsx
@@ -1,6 +1,6 @@
 import { action } from 'storybook/actions';
 
-import { PharosCheckbox, PharosLink } from '../../react-components';
+import { PharosCheckbox, PharosCheckboxGroup, PharosLink } from '../../react-components';
 import { defaultArgs, type ComponentArgs, type StoryArgs } from './storyArgs';
 import { configureDocsPage } from '../../utils/_storybook/docsPageConfig';
 import { PharosContext } from '../../utils/PharosContext';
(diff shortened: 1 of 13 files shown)


packages/pharos/src/components/checkbox/PharosCheckbox.react.stories.tsx
```diff
@@ -122,6 +122,65 @@ export const Validity: Story = {
...
+          <PharosCheckbox value="email" fullWidth checked>
+            <span slot="label">This is the first choice</span>
+          </PharosCheckbox>
+          <PharosCheckbox value="product" fullWidth>
+            <span slot="label">This is the second choice</span>
+          </PharosCheckbox>
+          <PharosCheckbox value="research" fullWidth>
+            <span slot="label">
+              This is the third choice with a label that is just entirely too long and someone
+              should have said something before shipping this to users
```
jmccullough-ithaka, line 175: 😂

brentswisher: Gotta demo those real life use cases 😄

phosphotungstic (approved):

Issue #1009, opened 2025-07-05
Migrate from @open-wc/testing to e.g. @testing-library


daneah:
**The problem**

Both @open-wc/testing and @open-wc/testing-helpers have not been published with regularity in the last couple of years. Recently, they have started degrading with the rest of the ecosystem, leading to the need to manage patches for them.

**The solution**

Move to a more widely-supported and current testing framework.

**Alternatives considered**

Worth looking into other available options, but Testing Library is a well-regarded favorite.

**Additional information**

Not to be confused with the parts we're using @web/test-runner for, we're just using @open-wc/testing for a few things specifically:

- `fixture`
- `expect`
- `aTimeout`
- `nextFrame`
- `elementUpdated`
- (probably accidentally since we usually import from `'lit/static-html.js'`, just a few instances of) `html`

brentswisher:
I'm not sure what all is needed for this change, but if we are going to need any significant migration effort on the unit tests I would love to at least look at replacing mocha and @web/test-runner too. 

They work, but it always takes a second to remember the mocha syntax which is different from all our other projects at ITHAKA, which use jest or vitest. I would love to compare and see if Vitest or [Vitest Browser Mode](https://vitest.dev/guide/browser/) would be a viable replacement for our entire current unit testing stack 😄 🫣

brentswisher:
Maybe this would be a good topic for the next working group swarm

Niznikr:
Vitest browser tests work great with Lit in my experience.

https://github.com/vitest-dev/vitest/tree/main/examples/lit
https://github.com/vitest-dev/vitest/blob/main/examples/lit/vite.config.ts

Issue #580, opened 2023-08-10
Unit test assertions about errors always pass


brentswisher:
**Expected behavior**
The unit tests would fail when invalid assertions are made about the errors being thrown

**Actual behavior**
The unit tests pass, even when making invalid assertions about the errors that are being thrown

**Steps to reproduce the issue**

1. Download Pharos Locally
2. Open the [pharos icon tests](https://github.com/ithaka/pharos/blob/develop/packages/pharos/src/components/icon/pharos-icon.test.ts) file
3. Change the "throws an error for an invalid icon name" assertion to be something silly:
```TypeScript
  it('throws an error for an invalid icon name', async () => {
    component = await fixture(html` <test-pharos-icon name="fake"></test-pharos-icon> `).catch(
      (e) => e
    );
    expect('Could not find the moon').to.be.thrown;
  });
```
4. Run tests using `yarn test`
5. See that the tests still pass

**Pharos version**
12.21.0 - latest `develop` branch

**Your environment**

- OS: macOS
- Browser: N/A
- Version: N/A

**Additional information**
It looks like this is just the wrong syntax when using chai, and it should be something like
```TypeScript
expect(badFn).to.throw('salmon');
```

While I found this in the `pharos-icon` component, a similar syntax is used in 18 other places in the library and should probably all be fixed
```
/pharos/packages/pharos/src/components/alert/pharos-alert.test.ts
  25,47:     expect('status is a required attribute.').to.be.thrown;

/pharos/packages/pharos/src/components/button/pharos-button.test.ts
  141,82:       expect('fake is not a valid type. Valid types are: button, submit, reset').to.be.thrown;

/pharos/packages/pharos/src/components/heading/pharos-heading.test.ts
  43,46:     expect('level is a required attribute.').to.be.thrown;
  50,82:     expect('7 is not a valid heading level. Valid levels are: 1, 2, 3, 4, 5, 6').to.be.thrown;
  59,7:     ).to.be.thrown;

/pharos/packages/pharos/src/components/icon/pharos-icon.test.ts
  21,39:     expect('Could not find the moon').to.be.thrown;

/pharos/packages/pharos/src/components/image-card/pharos-image-card.test.ts
  137,81:     expect('fake is not a valid variant. Valid variants are: base, collection').to.be.thrown;
  151,7:     ).to.be.thrown;
  165,7:     ).to.be.thrown;

/pharos/packages/pharos/src/components/layout/pharos-layout.test.ts
[...]

PR #1300, opened 2026-06-12
Migrate test runner to vitest browser 

### LINKED ISSUE #1169: Getting started guide does not cover playwright installation for running tests
The getting started guide does not mention that installing playwright via 

```shell
npx playwright install
```

is necessary for the test suite to run properly. The `yarn test` command will fail without running any tests, and an error message will appear saying the user should run `yarn playwright install` to install playwright - but that does not work in this repo setup. This causes confusion with how to resolve the issue.

brentswisher:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [X] Improves maintainability
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [ ] Changeset added?

Starts the process of migrating away from `@open-wc/testing` (#1009)

Our current testing framework has proven to be somewhat hard to work (#580) with and not as well supported ([current version](https://github.com/ithaka/pharos/blob/bee2c393e933ce573a8a0362c83447ee50da8bc0/packages/pharos/package.json#L71)) as we would like. While the desired end result is to get rid of `@open-wc/testing`, doing so all in one pass has proven to be difficult enough that it keeps getting delayed or dropped in favor of more pressing work.

This PR takes things from the other direction - replacing the current `wtr` test _runner_ with `vitest` using browser mode, but keeping the tests themselves largely the same. This allows us to start moving in the direction we want to go with minimal changes to the large testing suite we need to maintain.

Once we get this merged in we can start working through the test files component-by-component as we have time and removing `@open-wc/testing`, we could probably create per-component issues or a project to track that work. 

It also has the fringe benefit of making `playwright` a direct devDependency (closes #1169 ). With `wtr`, playwright was required but not listed as a dependency. This requiring contributors knowing what was happening and running `npx playwright install` locally to get tests to pass, which was a frequent stumbling block for new contributors.


packages/pharos/vitest.config.ts
```diff
@@ -0,0 +1,74 @@
+import { fileURLToPath } from 'node:url';
+import { defineConfig } from 'vitest/config';
+import { playwright } from '@vitest/browser-playwright';
+
+const silencedLogs = ['Lit is in dev mode.', 'Multiple versions of Lit loaded.'];
+
+// Pre-existing unhandled rejections that wtr ignored, which we can now see and are silencing until resolved.
```
brentswisher, line 7: We can add a follow-up issue to address these

packages/pharos/src/components/image-card/pharos-image-card.test.ts
```diff
@@ -812,7 +811,6 @@ describe('pharos-image-card', () => {
     );
 
     component.focus();
-    sendKeys({ press: 'Tab' });
```
brentswisher, line 815: This was never actually doing anything meaningful, I removed just this line on `develop` as well and the test passes just the same.

packages/pharos/src/components/pill/pharos-pill.test.ts
brentswisher, line 1: These tests passed intermittently as they were when I was running them locally, extending the timeout made them less flaky

packages/pharos/src/components/checkbox-group/pharos-checkbox-group.test.ts
```diff
@@ -142,7 +142,9 @@ describe('pharos-checkbox-group', () => {
       .querySelector('fieldset')
       ?.getAttribute('aria-describedby');
 
-    expect(component.renderRoot.querySelector(`#${groupDescID}`)?.textContent)?.to.equal(text);
+    expect(component.renderRoot.querySelector(`#${groupDescID}`)?.textContent?.trim())?.to.equal(
```
brentswisher, line 145: Because of the difference in how vitest compiles the component for testing, there were some whitespace changes, which resulted in errors like this
```
AssertionError: expected '\n            Please make a selection\n            \n          ' to equal 'Please make a selection'
```

I just added `.trim()` where this was happening

packages/pharos/vitest.config.ts
```diff
@@ -0,0 +1,74 @@
...
+        functions: 96,
+        lines: 94,
+      },
+    },
+    browser: {
+      enabled: true,
+      headless: true,
+      provider: playwright(),
+      screenshotFailures: false,
+      viewport: { width: 1280, height: 720 },
```
brentswisher, line 70: Match wtr's full-page size, Vitest iframe default is 414x896

daneah (approved):

satya-achanta-venkata (approved):

michael-iden (approved):

Issue #1331, opened 2026-07-08
No error is emitted when `selected` attribute set on a non-selectable variant of the image card


brentswisher:
**Expected behavior**
If I somehow accidentally pass `selected` to an image card that is not one of the selectable variants, it should emit an error message alerting me

**Actual behavior**
No error message is emitted, but a visual border is added around the image card, breaking the expected design

**Steps to reproduce the issue**

1. Create an image card with the default variant
2. Add "selected" to the component
3. See the broken design
4. See no error was emitted to the console

**Screenshots or code**
<img width="1249" height="445" alt="Image" src="https://github.com/user-attachments/assets/57796960-ce38-467b-9058-5f9647703413" />

**Pharos version**
v14.23.0

**Your environment**
- OS: macOS
- Browser: Chrome

**Additional information**
This looks to be the related logic: https://github.com/ithaka/pharos/blob/main/packages/pharos/src/components/image-card/pharos-image-card.ts#L181-L185C6

🤔 I actually think this is maybe both inverted and checking the wrong attribute? 
It should probably be something like `if (this._isSelected && !this._isSelectable())`

PR #1332, opened 2026-07-08
Migrate unit tests to Vitest

### LINKED ISSUE #1009: Migrate from @open-wc/testing to e.g. @testing-library

brentswisher:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Closes #1009 and #580

This updates all of the unit tests to use Vitest, by translating the assertions from open-wc/sinon to vitest.

**Additional context**

A few notes:
- The test fixture no longer clears the DOM between tests. Vitest's browser mode handles isolation between test files, and the manual clearing caused issues with some tests which built additional dom structure insinde the test, but outside the fixture. For tests where the DOM does need to be cleared between tests, it is now done in a afterEach block.
- There are a handful of places where errors are thrown in async functions because they use the `updated` lifecycle method. The `errorFixture` doesn't work for those case. I added one-off error listening/suppression in these few spots rather that update the fixture to handle them as doing so really complicated it.
- One image card test is now skipped, it surfaces a pre-existing bug filed in #1331
- Where tests previously relied on open-wc's `elementUpdated`/`waitUntil` for components that re-render asynchronously, those are now `vi.waitFor(() => expect(...))` calls that wait on the settled state rather than guessing frame counts or timing.


packages/pharos/src/components/button/pharos-button.test.ts
```diff
@@ -13,84 +14,86 @@ describe('pharos-button', () => {
     component = await fixture(html` <test-pharos-button>I am a button</test-pharos-button> `);
   });
 
+  afterEach(() => {
```
brentswisher, line 17: Vitest clears the DOM between test file, but not tests. Tests which reference the document directly need to include cleanup like this to prevent cross-test interference

packages/pharos/src/components/image-card/pharos-image-card.test.ts
```diff
@@ -136,49 +138,69 @@ describe('pharos-image-card', () => {
         <div slot="metadata">Description of collection.</div>
       </test-pharos-image-card>`
     );
-    await expect(component).to.be.accessible();
+    await expect(component).toBeAccessible();
   });
 
   it('throws an error for an invalid variant value', async () => {
```
brentswisher, line 144: This is arguable a bug in the component we should fix, but for now just wanted to limit changes to the tests

packages/pharos/src/components/sheet/pharos-sheet.test.ts
```diff
@@ -104,33 +106,39 @@ describe('pharos-sheet', () => {
...
     };
     document.addEventListener('focusin', onFocusIn);
 
     const trigger = document.createElement('button');
     trigger.setAttribute('id', 'trigger');
-    trigger.setAttribute('data-sheet-id', 'my-sheet');
+    trigger.setAttribute('data-sheet-id', 'focus-sheet');
     document.body.appendChild(trigger);
 
-    const button = document.querySelector('#trigger') as HTMLButtonElement;
```
brentswisher, line 122: I'm not sure why this was reselcting the trigger with a querySelector before?

packages/pharos/src/components/button/pharos-button.test.ts
lizlove: can you clarify what counts as 'referencing the document directly' here?

lizlove (approved):
Congrats ! 🥳 
Had a quick clarifying question but otherwise 👏

packages/pharos/src/components/button/pharos-button.test.ts
brentswisher: Yeah, no problem, at least this is my understanding 😄

In some of the button tests like this one:

https://github.com/ithaka/pharos/blob/260a8b99c9d3a6f0b391f41d54d45e3e6b6153c3/packages/pharos/src/components/button/pharos-button.test.ts#L295-L312

We are creating a form element, and passing it in to the fixture function as the parent element.

The fixture then mounts the component in that element and appends it to the page DOM using document.body.appendChild:

https://github.com/ithaka/pharos/blob/260a8b99c9d3a6f0b391f41d54d45e3e6b6153c3/packages/pharos/src/test/fixture.ts#L17-L24

The test then runs and asserts thing, but Vitest browser **won't clean up that form element from the DOM**. It only does that automatically in between different test files, not individual tests.

Then, in the next test we do the same thing: add a form element, mount, etc.

But when it gets to this line, the selector will find 2 form elements in the page DOM, one from the previous test and one from the current test, and fail.

https://github.com/ithaka/pharos/blob/260a8b99c9d3a6f0b391f41d54d45e3e6b6153c3/packages/pharos/src/components/button/pharos-button.test.ts#L314

Adding the afterEach block will clear the DOM between tests to keep that from happening.

PR #1343, opened 2026-07-21
Correctly throw error when un-selectable image card is marked as selected

### LINKED ISSUE #1331: No error is emitted when `selected` attribute set on a non-selectable variant of the image card

brentswisher:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Fixes #1331 

The image card has several variants which can be selected: (https://pharos.jstor.org/storybook/?path=/story/web-components_components-image-card--selectable)

This can be set programmatically by passing a `selected` attribute to the web component. The component has some logic that if `selected` was set on an image card but it was not one of the variants which allowed for selection, it should throw an error.

There was a bug in the logic, while the HTML attribute is `selected`, the Lit property that maps to it is  `_isSelected`, and the code was checking the wrong thing. In addition, the logic was inverse so if it was working it would have thrown errors for selectable variants when they were selected 🙃 .

This fixes the logic so the error will be thrown as originally intended, and re-enables the test that exercises this logic.

### DIFF
--- a/.changeset/thirty-poems-cough.md
+++ b/.changeset/thirty-poems-cough.md
@@ -2,4 +2,4 @@
 '@ithaka/pharos': patch
 ---
 
-Corretly throw an error when a non-selectable image card is marked as selected
+Correctly throw an error when a non-selectable image card is marked as selected
--- a/.changeset/thirty-poems-cough.md
+++ b/.changeset/thirty-poems-cough.md
@@ -1,0 +1,5 @@
+---
+'@ithaka/pharos': patch
+---
+
+Corretly throw an error when a non-selectable image card is marked as selected
--- a/packages/pharos/src/components/image-card/pharos-image-card.test.ts
+++ b/packages/pharos/src/components/image-card/pharos-image-card.test.ts
@@ -177,10 +177,7 @@ describe('pharos-image-card', () => {
     );
   });
 
-  // This test was discovered to be silently failing as part of https://github.com/ithaka/pharos/issues/1009
-  // Bug filed here https://github.com/ithaka/pharos/issues/1331
-  // We can un-skip this once that is resolved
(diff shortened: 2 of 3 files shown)


satya-achanta-venkata (approved):

.changeset/thirty-poems-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ithaka/pharos': patch
+---
+
+Corretly throw an error when a non-selectable image card is marked as selected
```
lizlove:
```suggestion
Correctly throw an error when a non-selectable image card is marked as selected
```

lizlove (changes requested):

.changeset/thirty-poems-cough.md
brentswisher: Thank you!

lizlove (approved):

PR #1344, opened 2026-07-21
chore(infra): add turbosnap

### LINKED ISSUE #514: Configure TurboSnap for faster Chromatic builds
**The problem**

Chromatic snapshots take a while. As we add C components and S stories, the snapshots take CxSxt time to run.

**The solution**

Chromatic provides [TurboSnap](https://www.chromatic.com/docs/turbosnap), a feature that detects changed components and takes snapshots of only changed components. This reduces time spent snapshotting per build, and also reduces our billed usage.

daneah:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [ ] Changeset added?

Resolves #514 

We have a lot of components. We are very cross-browser. We don't want to build so many things. TurboSnap makes it faster and cheaper to test Storybook by testing only changed components.

Add the config for TurboSnap to the Chromatic workflow

**Additional context**

[Docs](https://www.chromatic.com/docs/turbosnap/)


package.json
```diff
@@ -26,7 +26,7 @@
     "test-storybook:react": "test-storybook --ci --config-dir .storybook/react --url http://127.0.0.1:9001",
     "prestorybook:wc:dev": "yarn storybook:setup",
     "storybook:wc:dev": "yarn prestorybook:wc:dev && concurrently \"yarn:storybook:wc:start\" \"yarn:watch:styles\" --raw --kill-others",
-    "storybook:wc:build": "storybook build -c .storybook/wc -o .storybook-static/wc",
+    "storybook:wc:build": "storybook build --webpack-stats-json -c .storybook/wc -o .storybook-static/wc",
```
brentswisher: Is this so it knows what pieces have changed?

brentswisher (approved):

package.json
daneah: Yeah, I didn't know about this but the output from the first commit's failure of the Chromatic job directed me to add it. Locally run, it contains an array with a bunch of items like the following:

```json
{
  "modules": [
    ...,
    {
      "id": "./packages/pharos/src/components/combobox/pharos-combobox.wc.stories.ts",
      "name": "./packages/pharos/src/components/combobox/pharos-combobox.wc.stories.ts",
      "reasons": [
        {
          "moduleName": "/virtual:/@storybook/builder-vite/storybook-stories.js"
        }
      ]
    },
    ...
  ]
}
```

daneah:
Hmm, I would've hoped #1345 would trigger and catch a change to the alert 😖

daneah:
The issue with TurboSnap detecting changes is ultimately related to the fact that Vite can't properly infer the dependency tree because in story files we use the registered custom element tags (e.g. `<storybook-pharos-alert>`) instead of importing the `PharosAlert` from its module.

Possible avenues:

1. We could import the class and use it as `meta.component` in each story, but that may not be sufficient for page-level stories and also breaks automatic docgen.
1. We could import the files just for the side effects, but that is easy to forget/miss/update. 1. I started looking into using ESLint as a way to identify when a file has a custom element reference but not an import of the related component module, and that's when I ran into #1356.
1. We could do something in the Vite transform level to auto-insert `import`s, but the code is a bit hairy.

lizlove (approved):

PR #1383, opened 2026-08-11
Add DOM cleanup to tooltip test suite

### LINKED ISSUE #1382: Tooltip tests are flaky in CI
**Expected behavior**

The `pharos-tooltip` unit tests should pass consistently.

**Actual behavior**

The tooltip tests have been intermittently failing in CI:

```
FAIL |chromium| src/components/tooltip/pharos-tooltip.test.ts > pharos-tooltip
  > opens the first one on focus and then closes it upon hovering the second
AssertionError: expected true to be false // Object.is equality
 ❯ src/components/tooltip/pharos-tooltip.test.ts:145:29
```

I noticed this in a recent dependabot update, but I saw it once before too, I think that is just exercising the issue. When we migrated to Vitest in #1332 we missed this set of tests requiring DOM cleanup in an `afterEach` block. This means the  tooltip elements accumulate across the file, which causes issues when we query later against the DOM.

1. Run `yarn test` in CI, or run the full `pharos-tooltip.test.ts` file under load.
2. The test at `pharos-tooltip.test.ts:145` fails in Chromium.

**Proposed fix**

Add an `afterEach` like the other test suites that ran into this issue have.

**Pharos version**

14.25.0

**Your environment**

- OS: Failure is in CI
- Browser: Chromium

brentswisher:

- [ ] Adds a new feature
- [ ] Fixes a bug
- [X] Improves maintainability
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [X] No

**Is the:** (complete all)

- [X] Test suite(s) passing?
- [X] Code coverage maximal?
- [ ] Changeset added?

Closes #1382 

This updates the `afterEach` block to clean up the document body after each test, to prevent lingering DOM elements from causing issues with later tests.

### DIFF
--- a/packages/pharos/package.json
+++ b/packages/pharos/package.json
@@ -79,7 +79,7 @@
     "chokidar": "^5.0.0",
     "globby": "^16.2.2",
     "minify-html-literals": "^1.3.5",
-    "playwright": "^1.61.1",
+    "playwright": "^1.62.1",
     "postcss": "^8.5.25",
     "prettier": "^3.9.6",
     "react": "^18.3.1",
--- a/packages/pharos/src/components/tooltip/pharos-tooltip.test.ts
+++ b/packages/pharos/src/components/tooltip/pharos-tooltip.test.ts
@@ -16,6 +16,9 @@ describe('pharos-tooltip', () => {
     trigger.setAttribute('id', id);
     trigger.setAttribute('data-tooltip-id', tooltipId);
     trigger.textContent = 'I am a button';
+    // Keep the trigger clear of the headless cursor at (0, 0), which would
+    // otherwise fire an unwanted mouseenter.
+    trigger.style.margin = '50px';
     document.body.appendChild(trigger);
     return trigger;
   };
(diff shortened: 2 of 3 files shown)


satya-achanta-venkata (approved):

daneah (approved):

brentswisher:
I spent some time this week trying to wrap my head around why exactly moving the button fixed things, just sharing here in case it's useful later. 

The test suite appends the trigger buttons as the first elements in document.body, which lands them in the top-left corner, and the headless browser's virtual cursor sits on it at (0, 0). So the browser fires a real `mouseenter` on the first trigger that the test never expected. That unexpected hover is what started breaking the "opens the first one on focus and then closes it upon hovering the second" test. It wasn't expecting the first button to be hovered, so both tooltips end up with a pending call to open a tooltip that created a race condition for which would open.

The Playwright bump included a Chromium update which seems to have changed something about the timing of how those event resolve, leaving the initial button hovered when we are expecting the second one. By moving the button we ensure the hover/focus events are happening as we intended in the test.

PR #1357, opened 2026-07-25
fix(sheet): fix broken Sheet component and add missing component documentation pages

satya-achanta-venkata:
**This change:** (check at least one)

- [x] Adds a new feature
- [x] Fixes a bug
- [ ] Improves maintainability
- [x] Improves documentation
- [ ] Is a release activity

**Is this a breaking change?** (check one)

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Title of this pull request clear, concise, and indicative of the issue number it addresses, if any?
- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

**What does this change address?**

Fixes #1089 — the Sheet component's docs page button did nothing when clicked.

While auditing the docs site for other gaps, we also found 4 additional components with Storybook stories but no documentation page: Checkbox group, Radio group, Popover, and Table. All 5 pages were confirmed missing on production (pharos.jstor.org returns 404 for each).

PR #1387, opened 2026-08-14
fix(popover): add background-color so content behind it no longer shows through

### LINKED ISSUE #1386: Popover component has no background-color, so content behind it shows through
**The problem**

While auditing the docs site for #1089 (missing Sheet documentation), we noticed the Popover's example on `/components/popover` visually overlapped the adjacent 'See live code examples in Storybook' link when opened (see PR #1357, [comment](https://github.com/ithaka/pharos/pull/1357#issuecomment-5294846615) and [reply](https://github.com/ithaka/pharos/pull/1357#issuecomment-5294851973)).

Investigating further with Chrome DevTools confirmed this isn't a z-index/stacking issue — the popover genuinely paints above surrounding content per the DOM. The real cause is that `.popover` in `packages/pharos/src/components/popover/pharos-popover.scss` has **no `background-color` at all**, only a `box-shadow`:

```scss
.popover {
  ...
  box-shadow: var(--pharos-elevation-level-4);
  border-radius: var(--pharos-radius-base-standard);
  ...
}
```

Confirmed via computed styles: `backgroundColor: rgba(0, 0, 0, 0)` (fully transparent). Because there's no opaque surface, anything visually positioned behind an open popover shows through it — this affects any real app using Popover, not just the docs site.

This likely traces back to #664 (Apply Elevation 4 to Popover), which added the `box-shadow` but not an accompanying background color.

**Comparison with sibling overlay components** (all of which set an explicit background):
- Dropdown menu: `background-color: var(--pharos-color-ui-10)`
- Tooltip: `background-color: var(--pharos-tooltip-color-background-base)`
- Modal content: `background-color: var(--pharos-modal-color-background-content)`
- Sheet content: `background-color: var(--pharos-color-white)`
- Popover: **none**

**The solution**

Add `background-color: var(--pharos-color-white);` to `.popover` in `pharos-popover.scss`, matching the pattern used by other elevated/floating surfaces (e.g. multiselect-dropdown, image-card, coach-mark) that pair `box-shadow: var(--pharos-elevation-level-*)` with `background-color: var(--pharos-color-white)`.

Will be addressed in a dedicated branch/PR separate from #1357, since this is a real `@ithaka/pharos` component defect (not a docs-only issue) with its own changeset/versioning implications.

satya-achanta-venkata:

- [ ] Adds a new feature
- [x] Fixes a bug
- [ ] Improves documentation
- [ ] Is a release activity

- [ ] Yes
- [x] No

**Is the:** (complete all)

- [x] Test suite(s) passing?
- [x] Code coverage maximal?
- [x] Changeset added?

Fixes #1386. The `.popover` class only set `box-shadow` (elevation) and never set an explicit `background-color`, leaving the computed background fully transparent (`rgba(0, 0, 0, 0)`). Because of this, any content behind the popover was visible through it, creating the illusion that the popover was rendering underneath other content even though its z-index and stacking order were correct.

- Adds `background-color: var(--pharos-color-white)` to `.popover` in `pharos-popover.scss`, matching the pattern already used by other elevated/floating components (Sheet, Modal, Dropdown menu, Tooltip).
- Adds a changeset documenting the patch.

**Additional context**
None.


jialin-he (approved):

brentswisher (approved):

satya-achanta-venkata:
Alright, considering we introduced background color, I reverted this change as [per the comment from Dane here](https://github.com/ithaka/pharos/pull/1357#issuecomment-5294851973). This is how popover renders now

Screenshot 2026-08-14 at 2 59 08 PM

